PR #4, opened 2021-05-17
Search Synthesis

edyounis:


examples/simple_leap.py
```diff
@@ -0,0 +1,70 @@
+"""This script is contains a simple use case of the QSearch synthesis method."""
```
WolfLink, line 1: typo in the comment “is contains”

examples/simple_qpredict.py
```diff
@@ -0,0 +1,29 @@
+"""This script is contains a simple use case of the QFAST synthesis method."""
```
WolfLink, line 1: This comment says QFAST instead of QPredict

WolfLink (commented):
This looks great!  This is really only missing three features from Qsearch:
- parallelism
- topology support in the layer generator
- variable gate weight support in the heuristics

Other than that it looks awesome, and is definitely improves upon the original qsearch implementation.

PR #11, opened 2021-05-23
QPredict

edyounis:


buttercutter:
Just a quick question : I am wondering if there are any citation or publication papers on [QPredict](https://github.com/search?q=org%3ABQSKit%20qpredict&type=code)

I could not find any in the [coding documentation](https://bqskit.readthedocs.io/en/latest/source/autogen/bqskit.passes.QPredictDecompositionPass.html) either.

@edyounis @ethanhs

edyounis:
Unfortunately, not. QPredict was never written up since it's a bit hit or miss at times. If you do use this tool and would like to cite it, you can use the [BQSKit reference](https://www.osti.gov/biblio/1785933).

Issue #17, opened 2021-05-25
Create CircuitLocation and CircuitRegion objects


edyounis:
CircuitLocation describes a set of qudits, and CircuitRegion describes a connected portion of a circuit. See https://github.com/BQSKit/bqskit/pull/11/files#r638133051 for the original discussion.

PR #23, opened 2021-05-26
[WIP] Iterative Reoptimization

edyounis:
Implementing several control flow and utility passes necessary for the iterative reoptimization project. Also implementing the reoptimization passes and updating the entire partitioning system.


edyounis:
Fixes #29
Implements #17

bqskit/compiler/passes/partitioning/scan.py
```diff
@@ -0,0 +1,180 @@
...
+            or model.num_qudits < circuit.get_size()
+        ):
+            _logger.warning(
+                'MachineModel not specified or invalid;'
+                ' defaulting to all-to-all.',
+            )
+            model = MachineModel(circuit.get_size())
+
+        # Find all connected, `block_size`-sized groups of qudits
+        # NOTE: This assumes circuit and topology qudit numbers are equal
```
mtweiden, line 118: Is this note necessary? Does the partitioner still work if we're using a `MachineModel` that has more qudits than the circuit? This is the usual use case for doing the hybrid topology runs, and its seems that expanding the circuit so that it fits the size of the coupling graph might hurt performance (more locations to check).

bqskit/compiler/passes/partitioning/scan.py
```diff
@@ -0,0 +1,180 @@
...
+        qudit_groups = model.get_locations(self.block_size)
```
mtweiden, line 119: It might be worthwhile to add a pruning function that removes qudit groups that do not have any qudits in the `used_qudits` set. That way the partitioner can perform well in cases where a very small circuit is partitioned on a very large `MachineModel`.

bqskit/compiler/machine.py
```diff
@@ -1,50 +1,51 @@
...
+    The MachineModel Class.
+
+    A MachineModel models a quantum processing unit's connectivity.
+    """
 
     def __init__(
         self,
         num_qudits: int,
-        coupling_graph: set[tuple[int, int]] | None = None,
+        coupling_graph: Iterable[tuple[int, int]] | None = None,
```
ethanhs, line 24: I think coupling graphs should support indexing, so they should be sequences.

bqskit/compiler/passes/partitioning/cluster.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+    The ClusteringPartitioner Pass.
+
+    This pass forms partitions in the circuit by placing points in the circuit
+    clustering the gates around them.
+    """
+
+    def __init__(self, block_size: int = 3, num_points: int = 8) -> None:
+        """
+        Construct a SimplePartitioner.
```
ethanhs, line 26:
```suggestion
        Construct a ClusteringPartitioner.
```

bqskit/compiler/passes/partitioning/scan.py
```diff
@@ -0,0 +1,216 @@
...
+    """
+
+    def __init__(
+        self,
+        block_size: int = 3,
+        single_gate_score: int = 1,
+        multi_gate_score: int = 1000,  # TODO: Pass callable scoring_fn instead
+    ) -> None:
+        """
+        Construct a SimplePartitioner.
```
ethanhs, line 35:
```suggestion
        Construct a ScanPartitioner.
```

Also do we want to call this the QGoPartitioner pass?

bqskit/compiler/passes/util/inject.py
```diff
@@ -0,0 +1,40 @@
...
+from typing import Any
+
+from bqskit.compiler.basepass import BasePass
+from bqskit.ir.circuit import Circuit
+
+
+_logger = logging.getLogger(__name__)
+
+
+class InjectDataPass(BasePass):
```
ethanhs, line 14: Is the idea behind this pass that people would add this part way through? Also perhaps this could be named `UpdateDataPass` or something?

bqskit/ir/point.py
```diff
@@ -15,5 +20,61 @@ class CircuitPoint(NamedTuple):
     cycle: int
     qudit: int
 
+    # TODO: Maybe add below code,
+    # so we can do CircuitPoint(point) instead of CircuitPoint(*point)
+
+    # def __new__(
+    #     cls,
+    #     cycle_or_tuple: int | tuple[int, int],
```
ethanhs, line 28: I think this would be good, would also be good to provide overloads here.

ethanhs (approved):
There's a bit too much for me to do a detailed review but I went over everything as best I could. I think designing the pass "metalayer" will be important to get right from a user perspective and I want to think more about that.

mtweiden (approved):

bqskit/compiler/passes/control/foreach.py
```diff
@@ -83,17 +83,16 @@ def run(self, circuit: Circuit, data: dict[str, Any]) -> None:
         # Perform work
         for point, op in blocks:
             gate: CircuitGate = op.gate  # type: ignore
-            circuit = gate._circuit.copy()
+            circuit = gate._circuit.copy()  # TODO: OVERWRITING CIRCUIT
```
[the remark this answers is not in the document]
edyounis: I think this is fine. I can't remember what the problem was here.

bqskit/compiler/passes/partitioning/scan.py
edyounis: In this case, I think the loop over qudit groups would cause a crash since the circuit isn't explicitly expanded. We should probably log a warning or error here.

bqskit/compiler/passes/partitioning/scan.py
edyounis: This should be handled by the construction of the divider.

bqskit/compiler/passes/partitioning/scan.py
```diff
@@ -0,0 +1,216 @@
...
+                        qudit_index: (divider[qudit_index], num_cycles)
+                        for qudit_index in qudit_group
+                    }),
+                )
+
+                in_qudits = list(q for q in qudit_group)
+                stopped_cycles = {q: num_cycles for q in qudit_group}
+                score = 0
+
+                for cycle, op in ops_and_cycles:
```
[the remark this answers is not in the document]
edyounis: I agree, but first, there is a lot of algorithm improvements to make.

bqskit/compiler/machine.py
edyounis: This is just what the user can pass in. It also isn't clear to me why the coupling graphs should support indexing. We are representing them as an edge set. Did you have a use case in mind?

bqskit/compiler/passes/partitioning/scan.py
edyounis: No, QGO refers to the entire partitioning-synthesis compilation algorithm. We should decouple that from the partitioner.

bqskit/compiler/passes/util/inject.py
edyounis: Yeah, basically there to give users more control over the execution of the passes. I like that name much better, will update shortly.

Issue #26, opened 2021-05-29
Fold method gives wrong unitary


mtweiden:
On the most up to date `topology-dev` branch, the fold method sometimes gives the wrong unitary. The minimal qasm code that produces this error is in the file `part13` and contains:
```
OPENQASM 2.0;
include "qelib1.inc";
qreg q[10];
cx q[0], q[8];
u3(0,0,-pi/512)  q[8];
cx q[0], q[8];
u3(0,0,pi/1024)  q[0];
u3(0,0,pi/512)  q[8];
cx q[1], q[8];
```
The error can be seen by taking the norm of the original circuit and the partitioned circuit using this code:
```
from bqskit.compiler.machine import MachineModel
from bqskit.compiler.passes.simplepartitioner import SimplePartitioner
from bqskit.ir.lang.qasm2.qasm2 import OPENQASM2Language
from numpy.linalg import norm

simp_part = SimplePartitioner(MachineModel(10), 3)
data = {}

with open('part13', 'r') as f:
    part13 = OPENQASM2Language().decode(f.read())
old_part13 = part13.get_unitary().get_numpy()
simp_part.run(part13, data)
new_part13 = part13.get_unitary().get_numpy()
print(norm(old_part13 - new_part13))
```
The expected outcome is that the norm is `0.0`.
The observed outcome is `39.1918...`.

PR #27, opened 2021-05-30
Fixed batch_pop bug

mtweiden:
The `qudits` list in the `batch_pop` method was unsorted, so occasionally when inserting a CircuitGate, the wrong qudits were mapped to. Solves the issue found in KEYWORD #26


edyounis (approved):

edyounis:
Nice! Good catch and fix, thanks!

PR #30, opened 2021-06-09
Update mypy and improve config

ethanhs:
This has some new nice things like `TypeGuard`

PR #33, opened 2021-06-12
Rust modifications

### LINKED ISSUE #31: Sympy adds about 1 second to start up time due to imports
We only use sympy to handle transpositions and permutations in the PermutationMatrix class. We should consider refactoring it out.

ethanhs:
These are all of the changes needed to get BQSKit to work with the Rust backend. I have manually tested and verified that Rust `CostFunction`s, Rust  `Minimizer`s, and Python `CostFunction`s with Rust `Minimizer`s all work. Right now the gate set is a little limited, but I will work on that next (it is the cause of the only failing test at the moment).

Things to do before merge (in no particular order):

- [x] export the needed permutation primitives to remove sympy from bqskit
- [x] publish bqskitrs on pypi (the tests fail otherwise)
- [x] write tests for the Rust CostFunctions
- [x] write tests for the Rust Minimizers
- [x] Make Ceres the default Minimizer
- [x] Implement all of the parametrized gates.
  - [x] U8
  - [x] RXX
  - [x] RYY
  - [x] RZZ
- [x] Support Python gates
- [x] Test all of the gates
- [x] Test the gradients a la https://github.com/BQSKit/qsearch/blob/master/tests/test_gradients.py

Fixes #31


ethanhs:
This should be ready for review now. Please try it out and report back!

.pre-commit-config.yaml
```diff
@@ -73,7 +73,7 @@ repos:
     rev: v0.800
     hooks:
     - id: mypy
-      additional_dependencies: [numpy>=1.20]
+      additional_dependencies: ["numpy>=1.20,<1.21"]
```
ethanhs, line 76: numpy 1.21 made `ndarray` generic, so we will need to update to fix that. I'll do those changes in #30

bqskit/ir/opt/instantiaters/qfactor.py
```diff
@@ -22,261 +20,21 @@
...
-
-            # Update current gate
-            if op.gate.is_parameterized():
-                env = ct.calc_env_matrix(op.location)
-                op.params = op.gate.optimize(env)  # type: ignore
-                # TODO: slowdown_factor
-
-            # Add updated gate to right of circuit tensor
-            ct.apply_right(op.get_unitary(), op.location)
+        return super().instantiate(circuit, target, x0)
```
ethanhs, line 37: Reminder to myself to restore the doc comment here.

setup.cfg
```diff
@@ -33,10 +33,10 @@ project_urls =
 [options]
 packages = find:
 install_requires =
+    bqskitrs==0.1.0_beta.1
```
ethanhs, line 36: Once this is tested a bit more I will bump this to a non-beta release.

bqskit/ir/opt/cost/functions/residuals/hilbertschmidt.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class HilbertSchmidtResidualsGenerator(CostFunctionGenerator):
+    """
+    The HilbertSchmidtResidualsGenerator class.
+
+    This generator produces configured HilbertSchmidtResiduals functions.
+    """
+
+    def gen_cost(
```
edyounis, line 39: Should we split the generators? We could allow the generator to generate a residual with `gen_residual` and add a `ResidualFunctionGenerator` base class?

bqskit/ir/opt/instantiaters/qfactor.py
```diff
@@ -22,151 +20,13 @@
...
-
-        self.diff_tol_a = diff_tol_a
-        self.diff_tol_r = diff_tol_r
-        self.dist_tol = dist_tol
-        self.max_iters = max_iters
-        self.min_iters = min_iters
-        self.slowdown_factor = slowdown_factor
-        self.reinit_delay = reinit_delay
-        self.log_delay = log_delay
+    def __new__(cls, **kwargs: dict[str, Any]) -> Any:
```
edyounis, line 26: Is there a way to handle at a higher level, like in `circuit.minimize`?

ethanhs: Probably, I think we should still have the `__new__` but it should (at least for now) reject all non-empty kwargs.

bqskit/ir/opt/cost/functions/residuals/hilbertschmidt.py
ethanhs: Yeah, perhaps. Technically HilbertSchmidtResiduals *is* a cost function, and the current system means that we don't need a bunch of

```python
if optimizer == 'bfgs':
     cost_fn_gen.gen_cost(..)
else:
     cost_fn_gen.gen_residual(..)
```
around the code base.

PR #37, opened 2021-07-16
Fast scan

mtweiden:
The `heap_scan` partitioner uses a heap to maintain a list of the best scoring qudit groups. Only qudit groups that might have changed in the previous scan iteration are reevaluated. There is probably opportunity for further improvement by getting rid of qudit groups that might have changed but didn't because of idle qudits, but I'll leave that for another day.


bqskit/compiler/passes/partitioning/scan.py
```diff
@@ -180,59 +265,41 @@ def run(self, circuit: Circuit, data: dict[str, Any]) -> None:
...
-                        for qudit in qudit_group
-                        # This statement
-                        # if stopped_cycles[qudit] - 1 >= divider[qudit]
-                    })
+            # Heap partitioning
+            # Get best (up to date) block
+            while len(score_heap) > 0:
+                (best_score, group_index) = heapq.heappop(score_heap)
+                best_score *= -1
+                if best_score == block_list[group_index].score:
```
edyounis, line 273: Why is this check necessary? Can't you just check if the heapq is not empty and pop the top one?

bqskit/compiler/passes/partitioning/scan.py
```diff
@@ -180,59 +265,41 @@ def run(self, circuit: Circuit, data: dict[str, Any]) -> None:
...
+                (new_region, new_score) = self._form_region_and_score(
+                    block_list[group_index].qudit_group,
+                    divider,
+                    circuit,
+                )
+                # Update the block_list
+                block_list[group_index].region = new_region
+                block_list[group_index].score = new_score
+                # push to heap
+                heapq.heappush(score_heap, (-1 * new_score, group_index))
```
edyounis, line 302: Multiple inserts into a heap can be made more efficient by first building a heap of the elements to insert then merging the two heaps. You can use `heapq.merge` to the do the merging.

See [here](https://cs.stackexchange.com/questions/105086/an-algorithm-to-efficiently-insert-a-list-of-elements-into-a-binary-heap-bulk) for more info.

bqskit/qis/unitary/unitarymatrix.py
```diff
@@ -119,7 +119,8 @@ def get_distance_from(self, other: UnitaryLike) -> float:
         other = UnitaryMatrix(other)
         num = np.abs(np.trace(other.get_numpy().conj().T @ self.get_numpy()))
         dem = self.get_dim()
-        dist = np.sqrt(1 - ((num / dem) ** 2))
+        val = 1 - ((num / dem) ** 2)
+        dist = np.sign(val) * np.sqrt(np.abs(val))
```
edyounis, line 123: `val` should not be absoluted. If the value is negative, we should round up to zero.

`dist = np.sqrt(val)`
The return statement on the next line will round up to zero.

bqskit/compiler/passes/partitioning/scan.py
```diff
@@ -135,38 +220,38 @@ def run(self, circuit: Circuit, data: dict[str, Any]) -> None:
             for q in range(circuit.get_size())
         ]
 
+        # Create the block_list, member_dict, and score_heap
+        # `block_list` maintains the current region and score fore each
+        # `qudit_group`.
+        block_list: list[Block]
+        block_list = []
+        # `member_dict` has key: an qudit in the Circuit and value: a list of
```
edyounis, line 228: The documentation here could be a little clearer. For example, one way of rewording it may be: `member_dict maps XXX to XXX such that XXX`.

edyounis (changes requested):
Good use of a heapq to solve this problem more efficiently! Thanks for implementing this. There are a few changes I requested. 

Have you tested this on the benchmarks for your experiments? Did you observe a significant speedup?

bqskit/compiler/passes/partitioning/scan.py
```diff
@@ -135,38 +220,38 @@ def run(self, circuit: Circuit, data: dict[str, Any]) -> None:
...
+        # indices into the `block_list` whose `qudit_group` contains the qudit
+        # key.
+        member_dict: dict[int, list[int]]
+        member_dict = {q: [] for q in range(circuit.size)}
+        # `score_heap` is a "max" heap that keeps track of the best scores, and
+        # the index into the `block_list` that has that score.
+        score_heap: list[tuple[int, int]]
+        score_heap = []
+
+        # TODO: Support avoiding operations that are too big to partition.
```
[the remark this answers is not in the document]
mtweiden: Just forgot to get rid of this note to myself.

bqskit/compiler/passes/partitioning/scan.py
mtweiden: Because I don't clear of out of date scores during the rescoring. This makes sure we're not using an out of date score to select the next best region.

bqskit/compiler/passes/partitioning/scan.py
mtweiden: Thanks for pointing that out, I've switched to this.

bqskit/compiler/passes/partitioning/scan.py
mtweiden: Reworded this.

bqskit/qis/unitary/unitarymatrix.py
mtweiden: Changed this back. I switched it to this in the first place because numpy was giving warnings about square rooting negatives.

mtweiden:
I tested it on the 15 qubit QFT with blocksize 5. With the old scan partitioner it took 11:06.479, with the new it took 9:25.854. So not the speed up we were hoping for (at least for smaller circuits), but definitely a bit better.

bqskit/compiler/passes/partitioning/scan.py
edyounis: Is not clearing them worth it? Clearing them could be done in logarithm time since the array is sorted and we know the scores of the old ones. Not clearing them, leads to increasing heap size. Every time a block is formed, we have to add probably hundreds of elements to the heap. Imagine a large circuit with tens of thousands to millions of blocks, this can easily lead to an out-of-memory error.

bqskit/qis/unitary/unitarymatrix.py
edyounis: Yeah that makes sense, we could round up to zero before the sqrt; gives the same effect but no warnings.

edyounis:
Have you profiled it to see where the time is going? It would be nice to look at a side-by-side comparison of pyinstrument reports between new and old. I would expect a more significant speedup.

bqskit/compiler/passes/partitioning/scan.py
mtweiden: I think clearing the old scores would still take linear time (because it's just a heap) but you're probably right about clearing it. Do you think it might be worth changing to a BST? Getting the max score would take O(log n) instead of O(1), but cleaning the old scores out would take O(k log n) instead of O(k n).

mtweiden:
I'm still getting numbers for the old one, but it looks like this matches the older pyinstrument outputs in that a good chunk of time is going to `union` and just stepping through the circuit.

Screen Shot 2021-07-17 at 4 27 48 PM

mtweiden:
Here's the old scan, it looks like the speed up in many cases is even less than what I measured before:
Screen Shot 2021-07-17 at 5 48 22 PM

bqskit/compiler/passes/partitioning/scan.py
edyounis: If you remember the old scores, you can find them in O(log n) time in the heap. Removing an element in the heap also takes O(log n) time. This process can be repeated for each of the k elements you would like to remove, so O(k log n).

edyounis:
It seems like the `model.get_locations` is returning too many items. So the number of `qudit_group`s is larger than it should be. I added a failing test for this in `tests.compiler.test_machine.py`.

PR #43, opened 2021-08-03
Fixed bug that selected small regions first

mtweiden:
There's no need to figure out the merge stuff, the only thing wrong was that the list of `potential_regions` was sorted least to greatest, and we were picking the smaller regions first instead of the larger ones.


edyounis (approved):
Nice catch!

PR #44, opened 2021-08-04
Fixed bounding_region bug

mtweiden:
The `bounding_region` for the subsequent calls of `surround` was using an upper bound that was too low, so just single points were being picked out.


edyounis (approved):
Again, nice catch! Thanks for reviewing.

PR #51, opened 2022-01-05
Implemented SeedLayerGenerator

edyounis:


bqskit/passes/search/generators/seed.py
```diff
@@ -0,0 +1,127 @@
...
+
+        if not isinstance(target, (UnitaryMatrix, StateVector)):
+            raise TypeError(
+                'Expected unitary or state, got %s.' % type(target),
+            )
+
+        if target.dim != self.seed.dim:
+            raise ValueError('Seed dimension mismatch with target.')
+
+        data['seed_seen_before'] = [self.hash_structure(self.seed)]
```
ethanhs, line 75: Perhaps we should make this a set and implement `__hash__` on `Circuit`?

bqskit/passes/search/generators/seed.py
```diff
@@ -0,0 +1,127 @@
...
+        filtered_successors = []
+        for s in successors:
+            h = self.hash_structure(s)
+            if h not in data['seed_seen_before']:
+                data['seed_seen_before'].append(h)
+                filtered_successors.append(s)
+
+        return filtered_successors
+
+    def hash_structure(self, circuit: Circuit) -> int:
```
ethanhs, line 121: We should probably test this to make sure it doesn't run into any collisions.

ethanhs (approved):
Looks pretty good, I think moving the hashing to `Circuit` so it can be used by other layer generators would be beneficial.

bqskit/passes/search/generators/seed.py
edyounis: Yeah, I'll make it a set.

I am hesitant to add a `__hash__` to `Circuit` since it is a mutable object. I use hashes here because the `Circuit`s here are representing circuit structures and are not going to change.

ethanhs: Good point! In that case I think this is fine.

ethanhs (approved):

PR #64, opened 2022-04-08
Instantiater Enchancements

edyounis:
Here are the action items after our meeting last week:

- [x] do away with `batched_instantiate`
- [x] Set 1 thread per worker as default
- [x] Avoid building error by caching target unitary
- [x] Separate instantiation starting point generation from instantiation
- [x] Add an error if there is a mismatch between the number of generated and expected instantiation starting points
- [x] Add support for 4-param block

- [ ] ~~Implement `get_all_results` for instantiation~~
*This one won't work cleanly with the updated instantiation pipeline. This is a also very niche use case, where an `Instantiater` can be used instead of `circuit.instantiate`.


bqskit/ir/circuit.py
```diff
@@ -2209,9 +2215,12 @@ def get_unitary_and_grad(
     def instantiate(
         self,
         target: StateLike | UnitaryLike,
-        method: str | None = None,
+        method: str | Instantiater | None = None,
         multistarts: int = 1,
         seed: int | None = None,
+        start_gen: MultiStartGenerator = RandomStartGenerator(),
+        score_fn_gen: CostFunctionGenerator = HilbertSchmidtCostGenerator(),
```
ethanhs, line 2222:
```suggestion
        cost_fn_gen: CostFunctionGenerator = HilbertSchmidtCostGenerator(),
```
To be consistent with naming elsewhere.

bqskit/ir/circuit.py
```diff
@@ -2209,9 +2215,12 @@ def get_unitary_and_grad(
     def instantiate(
         self,
         target: StateLike | UnitaryLike,
-        method: str | None = None,
+        method: str | Instantiater | None = None,
         multistarts: int = 1,
         seed: int | None = None,
+        start_gen: MultiStartGenerator = RandomStartGenerator(),
```
ethanhs, line 2221:
```suggestion
        multistart_gen: MultiStartGenerator = RandomStartGenerator(),
```

(You'll need to change its usage below too)

ethanhs (approved):
Just a couple of variable name suggestions, otherwise LGTM

bqskit/ir/circuit.py
edyounis: We actually can't name it that, because we use it elsewhere. If we want the ability to specify two cost functions, one for use during instantiation, and one for use to compare instantiation results, we need two names. I had it originally called `cost_fn_gen`, but the instantiaters were complaining that they weren't getting a `cost_fn_gen` since that usually gets captured in the `**kwargs` and passed to the instantiater.

ethanhs: Ah got it.

PR #70, opened 2022-04-13
Quest Improvements

edyounis:


bqskit/exec/runners/quest.py
```diff
@@ -278,3 +295,81 @@ def annealing_objective(x: npt.NDArray[np.float64], *args: Any) -> float:
...
+
+    Args:
+        block_size (int): The maximum number of qudits in each
+            block. Defaults to 3.
+
+        compiler (Compiler | None): The compiler object used
+            to partitioning and synthesize the circuit. If left
+            as None, then a standard compiler will be created.
+
+        weit (float): The weight set on number of CNOTs versus
```
ethanhs, line 321: I would write out the full name `weight` here, its not much longer, and I've never seen `weit` before. Same with `pert` below.

ethanhs (approved):
Just a small suggestion about variable names, otherwise LGTM

bqskit/exec/runners/quest.py
edyounis: I agree with you, I was sticking with the vocab used in the prototype QuEST algorithm. I think we should keep the vocab matching, even if we aren't fans of it.

PR #72, opened 2022-04-15
Improve Multistart Generators

ethanhs:
This adds a `NaiveStartGenerator` which puts starting points along the diagonal of the search space.

We also expand the starting point distribution from [0, 1) to [0, 2*pi) so that we span the full space of the parameters.

Also fixes a couple of docstrings and the mypy config.

### DIFF
--- a/bqskit/ir/opt/multistartgens/naive.py
+++ b/bqskit/ir/opt/multistartgens/naive.py
@@ -1,4 +1,4 @@
-"""This module implements the RandomStartGenerator class."""
+"""This module implements the NaiveStartGenerator base class."""
 from __future__ import annotations
 
 from typing import TYPE_CHECKING
@@ -15,8 +15,8 @@ if TYPE_CHECKING:
     from bqskit.qis.unitary.unitarymatrix import UnitaryMatrix
 
 
-class RandomStartGenerator(MultiStartGenerator):
-    """A start generator that selects random points uniformily across [0, 1)"""
+class NaiveStartGenerator(MultiStartGenerator):
+    """A generator that puts starts along the diagonal of the N-d space."""
 
     def gen_starting_points(
         self,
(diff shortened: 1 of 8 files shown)


bqskit/ir/opt/multistartgens/random.py
```diff
@@ -16,6 +16,7 @@
 
 
 class RandomStartGenerator(MultiStartGenerator):
+    """A start generator that selects random points uniformily across [0, 1)"""
```
edyounis, line 19:
```suggestion
    """A start generator that selects random points uniformily across [0, 2π)"""
```

bqskit/ir/opt/multistartgens/naive.py
```diff
@@ -0,0 +1,63 @@
...
+from bqskit.utils.typing import is_integer
+
+if TYPE_CHECKING:
+    import numpy.typing as npt
+    from bqskit.ir.circuit import Circuit
+    from bqskit.qis.state.state import StateVector
+    from bqskit.qis.unitary.unitarymatrix import UnitaryMatrix
+
+
+class NaiveStartGenerator(MultiStartGenerator):
```
edyounis, line 18: I don't think this is very Naive, maybe we should call it DiagonalStartGenerator instead? Thoughts?

edyounis (commented):
Thanks for implementing this!

It might be beneficial to make this the default start generator for LEAP, since you have had previous success with it. This can be done in the default `instantiate_options` for the `LEAPSynthesisPass`.

Also, the documentation is built from one-level deep package `__init__.py` files, i.e. `bqskit.ir.__init__.py`, so we should include this in there. However, when I checked it out I left a TODO there to add more from the `opt` package. This is probably beyond the scope of this PR, but we should include the `opt` package in the documentation.

bqskit/ir/opt/multistartgens/naive.py
ethanhs: Sure! I mean it was naive in the sense that it was an educated guess of how to improve multistart. But I think diagonal is much more descriptive.

ethanhs:
I undid the changes to LEAP and Qsearch based on the data I found. So I think this is ready to merge.

edyounis (approved):

PR #77, opened 2022-04-18
Fix issue with mypy taking too long to run

ethanhs:
A clean `pre-commit run mypy --all-files` takes less than 15 seconds!


ethanhs:
I'm going to go ahead and merge this because it is simple and other CI is a lot slower because of it.

PR #78, opened 2022-04-18
Fix parallel Circuit.instantiate and passing a LocalCluster to Compiler

ethanhs:
This PR fixes two issues:

1) If you call `Circuit.instantiate` from the main thread the call to `secede` fails

2) If you try to pass a `LocalCluster` to `Compiler()` it will fail because it doesn't want us passing `silence_logs`.


bqskit/compiler/compiler.py
```diff
@@ -43,11 +44,14 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
             All arguments are passed directly to Dask. You can use
             these to connect to and configure a Dask cluster.
         """
-        dask_options = {
-            'silence_logs': logging.getLogger('bqskit').level,
-        }
-        if 'address' not in kwargs and 'scheduler_file' not in kwargs:
-            dask_options['threads_per_worker'] = 1
+        if any(isinstance(arg, LocalCluster) for arg in args):
```
edyounis, line 47: NIT: Should probably check if its an instance of the base class: `distributed.deploy.spec.SpecCluster`

edyounis (approved):
LGTM

PR #83, opened 2022-04-23
Implemented ExhaustiveGateRemovalPass

edyounis:


bqskit/passes/processing/iterative.py
```diff
@@ -0,0 +1,106 @@
...
+_logger = logging.getLogger(__name__)
+
+
+class IterativeScanningGateRemovalPass(PassAlias):
+    """
+    The IterativeScanningGateRemovalPass class.
+
+    Starting from one side of the circuit, attempt to remove gates one-by-one.
+    Repeat this until the circuit no longer changes. This will partition the
+    circuit first if the circuit width is too large.
```
ethanhs, line 32:
```suggestion
    circuit first if the circuit width is greater than a given width (5 qudits by default).
```

ethanhs (approved):
Looks good, just a couple of suggestions.

PR #91, opened 2022-08-09
Added maximal_matching function to graph, added supermarq stats to ci…

mtweiden:
circuit.py updated to include SupermarQ metrics (https://arxiv.org/abs/2202.11045).
graph.py updated to include method to find maximal matchings in coupling graphs.


edyounis (approved):
Thanks for the contribution!

PR #106, opened 2022-10-28
Batching using jax

alonkukl:


alonkukl:
We aren't going to merge to master anytime soon

PR #115, opened 2023-01-31
A faster implementation of the subgraph creation

alonkukl:
Changing the _adj data structure to hold a set of neighbors, and using it while creating the sub graph instead of iterating over all the edges of the graph.


edyounis:
Thanks for this! Nice catch on this huge performance bug. I will push this into the daskless branch since we need this now for some experiments :). It will officially be released in the 1.1 update. We can ignore the test failures for now as they are unrelated to your code changes.

PR #116, opened 2023-02-08
Added CCP gate, added BlockAnalysisPass

mtweiden:


bqskit/passes/util/blockanalysis.py
```diff
@@ -0,0 +1,80 @@
...
+    analyzed and compared quickly.
+
+    This pass scans through a circuit, and stores the results of the scan in a
+    list variable called `results`.
+    """
+
+    def __init__(
+        self,
+        analysis_function: Callable[[Circuit], Any] = gate_count,
+        filter_function: Callable[[Circuit], Any] = lambda x: x,
```
edyounis, line 35: Lambdas are not pickleable and will cause issues with the runtime system. You should define a module-level function to be used as a default here, similar to what you did for `gate_count`.

bqskit/passes/util/blockanalysis.py
```diff
@@ -0,0 +1,80 @@
...
+
+            filter_function (callable | None): A function that will filter out
+                blocks that meet some criteron. If `filter_function` returns
+                False for some block, it will be filtered out of the analysis.
+                It is assumed that this function takes a Circuit as an arugment.
+                (Default: lambda x: x)
+        """
+        self.analysis_function = analysis_function
+        self.filter_function = filter_function
+        self.results: list[Any] = []
```
edyounis, line 54: You shouldn't store any mutable data in the pass, rather, define a new list at the start of each `run` call.

bqskit/passes/util/blockanalysis.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def run(self, circuit: Circuit, data: dict[str, Any] = {}) -> None:
```
edyounis, line 56: You should switch the PR target branch to the new daskless branch and add an `async` keyword before the run here.

bqskit/passes/util/blockanalysis.py
```diff
@@ -0,0 +1,80 @@
...
+        """Perform the pass's operation, see :class:`BasePass` for more."""
+
+        if len(self.results) != 0:
+            _logger.debug('Clearing BlockAnalysisPass results...')
+            self.results = []
+
+        for op in circuit:
+            if type(op.gate) != CircuitGate:
```
edyounis, line 64: Rather than explicitly checking the type, you should `isinstance(op.gate, CircuitGate)`. This ensures that any custom subclassing logic is executed.

bqskit/passes/util/blockanalysis.py
```diff
@@ -0,0 +1,80 @@
...
+                continue
+
+            # Convert to subcircuit
+            block = Circuit(op.num_qudits)
```
edyounis, line 68: You can retrieve the `Circuit` in a `CircuitGate` object via `op.gate._circuit`.

bqskit/passes/util/blockanalysis.py
```diff
@@ -0,0 +1,80 @@
...
+                [_ for _ in range(op.num_qudits)],
+                op.params,
+            )
+
+            block.unfold_all()
+            if not self.filter_function(block):
+                continue
+
+            result = self.analysis_function(block)
+            self.results.append(result)
```
edyounis, line 80: I would think you want to store the results in the `data` dictionary at the end of the pass, so future passes can use this information. How are you currently using it?

edyounis (changes requested):
Thanks a lot for the contribution! There are a few small changes requested, but it looks awesome!

PR #127, opened 2023-03-13
Update Multistart API

edyounis:
Changed the instantiation API by moving the multi-start handling from `circuit.instantiate` to the `Instantiater`.

### DIFF
--- a/bqskit/ir/circuit.py
+++ b/bqskit/ir/circuit.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import copy
 import logging
+import warnings
 from typing import Any
 from typing import cast
 from typing import Collection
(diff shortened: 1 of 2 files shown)


bqskit/ir/circuit.py
```diff
@@ -2752,108 +2761,8 @@ def instantiate(
 
         instantiater = cast(Instantiater, instantiater)
 
-        # Check Target
-        try:
```
alonkukl, line 2756: Removing this casting is a big change, that breaks the API, are you sure you want to do that?

bqskit/ir/opt/instantiater.py
```diff
@@ -54,6 +56,42 @@ def instantiate(
...
+            circuit (Circuit): The circuit template to instantiate.
+
+            target (UnitaryMatrix | StateVector | StateSystem): The unitary
+                matrix to implement or state to prepare.
+
+            num_starts (int): The number of starting points to attempt
+                instantiation with.
+
+        Returns:
+            (Circuit): The paremeters for the circuit that makes the
```
alonkukl, line 78: It should be the circuit with the best parameters

alonkukl (changes requested):
I think that the API change of not converting the target is a big difference that should be talked over.

bqskit/ir/circuit.py
edyounis: I think the casting there was redundant as it was also cast inside Instantiater as well.

bqskit/ir/opt/instantiater.py
```diff
@@ -54,6 +56,42 @@ def instantiate(
...
+            Instantiater object may happen in parallel.
+        """
+        target = self.check_target(target)
+        start_gen = RandomStartGenerator()
+        starts = start_gen.gen_starting_points(num_starts, circuit, target)
+        cost_fn = HilbertSchmidtCostGenerator().gen_cost(circuit, target)
+        params_list = [self.instantiate(circuit, target, x0) for x0 in starts]
+        params = sorted(params_list, key=lambda x: cost_fn(x))[0]
+        circuit.set_params(params)
+        return circuit
```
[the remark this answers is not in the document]
edyounis: We want to return the same circuit here. That way, `circuit.instantiate(...)` and `circuit = circuit.instantiate(...)` mean the same thing, which is especially helpful when mapping `instantiate` calls through the runtime. I added a note mentioning this.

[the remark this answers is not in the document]
alonkukl: It seems strange that the function returns and changes in place.

PR #132, opened 2023-03-15
Bgate

mtweiden:
Added the B  and CCP gates. Added tests for B, CP, and CCP gates.


mtweiden:
The `BGate` and the `CCPGate` have been added to  `fill_gate_defs` function.

edyounis (changes requested):
Looks awesome! Thanks for working with all the boilerplate that goes into defining a gate. We should really see if there is a better way to handle new gates.

One last thing, I think you forgot to add CCPGate to the documentation in `bqskit.ir.gates.__init__`. After that, feel free to merge whenever. Thanks again!

PR #154, opened 2023-05-28
adding qudit gates

### LINKED ISSUE #153: Better Qudit Gate Support
The BQSKit IR has always supported qudit gates, but the library of gates included lacks some important ones. We should add these missing qudit gates to our library:

- Weyl-Heisenberg
- Generalized-Cliffords
- Subspace single-qudit gates
- Subspace swaps

We do include some qudit gates like the [CSUM](https://github.com/BQSKit/bqskit/blob/master/bqskit/ir/gates/constant/cpi.py) and [CPI](https://github.com/BQSKit/bqskit/blob/master/bqskit/ir/gates/constant/csum.py), which you can look to for an example.

khnikhil:
closes #153 as part of the UnitaryHack 2023 event. 

**NEW IMPLEMENTATIONS**:
1. a generic `QuditGate` base class (following the examples of `QubitGate` and `QutritGate`)
2. single-qudit Weyl-Heisenberg gates (`ShiftGate` and `ClockGate`)
3. a (Clifford) Hadamard gate `HDGate` (for an *n*-level qudit system)
4. qudit phase gates `PDGate` (related to $\sigma_z$ )
5. a generic (Clifford) CSUM gate `CSUMDGate` (following the [existing implementation in bqskit](https://github.com/BQSKit/bqskit/blob/master/bqskit/ir/gates/constant/csum.py), but generalized to qudits instead of just qutrits)
6. subspace swap gate `SubSwapGate`

Together, the generic `CSUMDGate`, `HDGate`, and `PDGate` s form a universal qudit gate set (proved in [this paper](https://pubs.aip.org/aip/jmp/article/56/3/032202/763827/Universal-quantum-computation-with-metaplectic)).

**FIXES**:
1. The existing `CSUMGate` had the wrong unitary representation (see screenshot directly below). I changed it to the correct unitary.

 Screen Shot 2023-05-28 at 9 49 50 PM

1. In #153, there were two other types of gates: subspace single-qudit gates, and subspace swaps. What is meant by "subspace" gates here? Does it refer to situations in which you have hybrid qubit-qudit systems?
3. Is my current workflow/placement of the new gates suitable? Since they are not parametrized, I put them in the `bqskit.ir.gates.constant` folder, but I don't use `ConstantGate` as a parent class. This is because the unitary matrix being implemented depends on the number of levels in the qudit (which is passed as an argument to the gate when it's called, and the unitary is built with a `get_unitary()` function). Is this a problem? Here is a screenshot of the code in practice.

Screen Shot 2023-05-28 at 1 02 38 AM

### DIFF
--- a/bqskit/ir/gates/constant/__init__.py
+++ b/bqskit/ir/gates/constant/__init__.py
@@ -18,6 +18,7 @@ from bqskit.ir.gates.constant.hd import HDGate
 from bqskit.ir.gates.constant.identity import IdentityGate
 from bqskit.ir.gates.constant.iswap import ISwapGate
 from bqskit.ir.gates.constant.itoffoli import IToffoliGate
+from bqskit.ir.gates.constant.pd import PDGate
 from bqskit.ir.gates.constant.permutation import PermutationGate
 from bqskit.ir.gates.constant.s import SGate
 from bqskit.ir.gates.constant.sdg import SdgGate
@@ -56,6 +57,7 @@ __all__ = [
     'IdentityGate',
     'ISwapGate',
     'IToffoliGate',
+    'PDGate',
     'PermutationGate',
     'SGate',
     'SdgGate',
--- a/bqskit/ir/gates/constant/__init__.py
+++ b/bqskit/ir/gates/constant/__init__.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 from bqskit.ir.gates.constant.ccx import CCXGate
 from bqskit.ir.gates.constant.ccx import ToffoliGate
+from bqskit.ir.gates.constant.clock import ClockGate
 from bqskit.ir.gates.constant.ch import CHGate
 from bqskit.ir.gates.constant.cpi import CPIGate
 from bqskit.ir.gates.constant.cs import CSGate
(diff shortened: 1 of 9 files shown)


edyounis:
Thanks for taking on this issue! Thanks for the awesome work as well.

You raise two questions answered here:
1. I am pretty sure you got the single-qudit ones. The subspace swaps are swaps between "qudit-levels" on a two-qudit gate. So for example, a |01> to |20> swap would be a 2-qudit gate that the identity with the |01> row and |20> rows swapped.
2. Yeah your workflow/placement is great!

khnikhil:
hi @edyounis,

I added a generic subspace swap gate `SubSwapGate` which can implement any swap between two qudit-levels, as you describe above (see below screenshot for a code example/verification). With that, I think that issue #153 has been completely addressed for the UnitaryHack event. Please let me know if there is anything else that needs to be done.

Screen Shot 2023-06-11 at 2 08 41 PM

edyounis:
Awesome! Once it passes the pre-commit we can merge this in and claim success. Thanks for your work on this!

khnikhil:
hi @edyounis, sounds great! I think this latest commit should fix the error, it seems I forgot to add type annotation for a helper function.

edyounis (approved):

PR #155, opened 2023-05-28
Add gate level get_inverse() method

### LINKED ISSUE #121: `DaggerGate` vs actually inverting the gate, qasm output for `DaggerGate`
The `get_inverse` method currently returns `DaggerGate` objects.  For small gates (single-qubit gates, CNOT, etc), is there a reason to save these as, eg, `DaggerGate(CNOTGate)` rather than just computing the adjoint (in this case, again CNOT)?

The rough use case is something like 
U^dg @ O @ U @ psi,
where U is given by some circuit.  

One might then want to print the full circuit, including reversed gates, to qasm (currently not possible with `DaggerGate`).  Also to get gate counts where `DaggerGate(CNOTGate)` is just counted as `CNOTGate`

Aaron-Robertson:
Fixes #121

### DIFF
--- a/bqskit/ir/circuit.py
+++ b/bqskit/ir/circuit.py
@@ -2491,13 +2491,7 @@ class Circuit(DifferentiableUnitary, StateVectorMap, Collection[Operation]):
         """Return the circuit's inverse circuit."""
         circuit = Circuit(self.num_qudits, self.radixes)
         for op in reversed(self):
-            circuit.append(
-                Operation(
-                    DaggerGate(op.gate),
-                    op.location,
-                    op.params,
-                ),
-            )
+            circuit.append(op.get_inverse())
         return circuit
 
     def get_unitary(self, params: RealVector = []) -> UnitaryMatrix:
--- a/bqskit/ir/gate.py
+++ b/bqskit/ir/gate.py
@@ -10,6 +10,8 @@ from typing import Callable
 from typing import ClassVar
 from typing import TYPE_CHECKING
 
+import numpy as np
+
 from bqskit.ir.location import CircuitLocation
 from bqskit.qis.unitary.unitary import Unitary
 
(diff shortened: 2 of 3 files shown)


Aaron-Robertson:
@edyounis here's the first pass on a get_inverse method. I have a few ideas on how to proceed, but would like some input!

Did you have a gate specific implementation in mind, or generalized as I started here? It certainly seems like the gate specific implementation is more in line with the codebase, in which the default DaggerGate would remain on the Gate base while specific implementations overwrite?

The current implementation would need at least specific cases for already implemented/known inverses (that are not self-inverse): e.g. T, S gates (to Tdg, Sdg), is the preference to store those mapping in a dict or on the gates themselves?

Finally in the parametrized case, I'm not sure exactly what's supposed to happen with the params. I thought about the FrozenParameterizedGate, but it didn't like the right choice looking over the implementation

Aaron-Robertson:
@WolfLink @edyounis Bump on the above comment if you have a moment! Still want to square away the preferred implementation (though on request I can just go for it 🚀🌔).

edyounis:
@Aaron-Robertson 

Thanks for taking this issue on! Sorry for the late reply, I have been on vacation.

> which the default DaggerGate would remain on the Gate base while specific implementations overwrite?

I think this is the way we want to go. Specifically, we can modify the `Circuit.get_inverse` method from:

```
    def get_inverse(self) -> Circuit:
        """Return the circuit's inverse circuit."""
        circuit = Circuit(self.num_qudits, self.radixes)
        for op in reversed(self):
            circuit.append(
                Operation(
                    DaggerGate(op.gate),
                    op.location,
                    op.params,
                ),
            )
        return circuit
```

to

```
    def get_inverse(self) -> Circuit:
        """Return the circuit's inverse circuit."""
        circuit = Circuit(self.num_qudits, self.radixes)
        for op in reversed(self):
            circuit.append(op.get_inverse())
        return circuit
```

then write the `Operation.get_inverse` method to use your new function `Gate.get_inverse` and a new function `Gate.get_inverse_params(self, params: RealVector) -> RealVector`. It would probably look something like:

```
    def get_inverse(self) -> Operation:
        """Return the operation's inverse operation."""
        return Operation(
            self.gate.get_inverse(),
            self.location,
            self.gate.get_inverse_params(self.params),
        )
```

I think the `is_self_inverse` check is awesome, and its cost is very acceptable because this will likely not be called from any performance-critical loop. The parameter check in the default implementation might not always be correct, we should isolate that logic to the gates where it is correct using the `Gate.get_inverse_params` function.

Aaron-Robertson:
Awesome @edyounis, thanks for the clear feedback, and hope your vacation was a blast! I am in the middle of a couple other issues, but I expect to have this one ready for another review tonight or tomorrow.

Aaron-Robertson:
@edyounis should we implement an abstract `get_inverse_params()` on `Gate` with a defined method on a subclass? I'm not quite seeing which subclass would do what we want, but I worry that a `ControlledGate` wouldn't fail on `get_inverse_params()`. For now, I'll push up the latest changes based on feedback!

edyounis:
The `get_inverse_params` default implementation should match the default implementation of the `get_inverse` method. So, I would leave a concrete implementation of `get_inverse_params` that returns the same params. This is correct if the inverse gate is a DaggerGate wrapper.

The logic you have about negating the params or flipping them around 2pi should be implemented in the gates where that logic is correct and should be accompanied by appropriate tests to verify that logic works well there.

Aaron-Robertson:
Alright @edyounis, thanks again for the feedback!

I decided to start implementing the tests, with a more general approach in place to identify which gates we should target with inverted params. That said, I'm wondering if this approach is suitable as it is, in which case we could be quite close. The idea is to generate the flipped-about-2pi params, and then simply check if the flip-parameterized gate is the inverse of the original, and to fall back to the dagger gate with original params if it isn't. What do you think?

nathanshammah:
@Aaron-Robertson if the PR is approved in the next few days (before next week starts), it will be considered for a bounty.

edyounis:
Hey @Aaron-Robertson, I hope you are doing well. I just wanted to touch base on this PR, please do let me know if you need any help or clarification. This is very close to being merge-ready.

edyounis:
Hey @Aaron-Robertson, I hope you're feeling better and doing well. I wanted to touch base on this PR. Do you think you will be able to finish it up soon?

edyounis:
Hey @Aaron-Robertson, I hope you are feeling better and doing well. Please let me know if you need any help with this PR, and I am happy to take it from here if you don't have any time to finish it up.

PR #170, opened 2023-08-29
Seeded synthesis

mtweiden:
Updated seeded synthesis pass for use in QSeed. Also created a structure analysis pass for easy profiling of partitioned circuits.


mtweiden:
Starting new pull request with seeded synthesis integrated into qsearch

PR #197, opened 2023-10-19
Support for Canonical Unitaries

mtweiden:
Adds the `canonical_unitary` function to `bqskit/utils/math.py`. Ensures that unitaries that differ only by a global phase are cast to the same matrix representation.

### DIFF
--- a/tests/utils/test_math.py
+++ b/tests/utils/test_math.py
@@ -10,6 +10,7 @@ import scipy as sp
 from scipy.stats import unitary_group
 
 from bqskit.qis.pauli import PauliMatrices
+from bqskit.utils.math import canonical_unitary
 from bqskit.utils.math import dexpmv
 from bqskit.utils.math import dot_product
 from bqskit.utils.math import pauli_expansion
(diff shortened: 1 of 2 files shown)


tests/utils/test_math.py
```diff
@@ -185,3 +186,28 @@ def test_valid(self, reH: npt.NDArray[np.complex128]) -> None:
         print(alpha)
         H = PauliMatrices(int(np.log2(reH.shape[0]))).dot_product(alpha)
         assert np.linalg.norm(H - reH) < 1e-16
+
+
+class TestCanonicalUnitary:
+    def random_phase(self) -> np.complex128:
```
edyounis, line 192: What happens if you go beyond 2pi range in phases? Since that's a huge chunk of the feature code, you should also test that here. You could also do property testing here with hypothesis.

edyounis (changes requested):

tests/utils/test_math.py
mtweiden: I didn't add property testing with hypothesis, but I've changed the test to use the `pytest.mark.parametrize` decorator. The random phase is now drawn according to a scaled normal distribution, so phases outside the range [0,2p) are also included.

edyounis (approved):
LGTM

PR #201, opened 2023-11-05
Squander decomposition

Lacika02:
I want to implement the Squander decomposition into qsearch.


edyounis:
I would love to see Squander's algorithms implemented inside of BQSKit! Thanks for taking the initiative here. I have to say, I have played around with the algorithms a lot previously, and having it cleanly callable from bqskit would be great.

I think the first question that needs to be answered here is whether we should rely on the Squander package or not. From a first glance, it seems that the squander package is not as portable as we would need it to be and requires manual compilation. Is this the case?

rakytap:
Dear [edyounis](https://github.com/edyounis),

We appreciate your positive response to the idea of implementing Squander algorithms into BQSKit. We have been exploring this idea for some time now and have initiated a project with my student to thoroughly examine this possibility in details.
At this point, I must mention that the creation of this PR occurred accidentally, as our development work is still in progress. Therefore, I propose closing this pull request after you have had the chance to review this message.

From a technical standpoint, I don't see any critical issues preventing us from incorporating SQUANDER as an optional extension to BQSKit. Installing SQUANDER from the PyPi repository is a straightforward process. The compilation of the C++ backends is automatically executed during the "pip install" process, provided that all dependencies are met. 
Recently we have switched to Apache 2.0 license by replacing all of the GPL licensed dependencies in our package.

I am optimistic that we can find suitable solutions for all the technical details, and I believe this initiative could lay the foundation for future collaborations.

Best regards,
Peter

edyounis:
> I am optimistic that we can find suitable solutions for all the technical details, and I believe this initiative could lay the foundation for future collaborations.

I believe the same! I will close this for now but look forward to hearing more from you and your team.

PR #208, opened 2023-11-30
Add reset

peachnuts:
Add reset operation and QASM support.


edyounis:
Sorry, I didn't mean to close this. It happened by accident when I merged and deleted 1.1.1. We need to update the base to be main, ping me on Slack, and we can figure it out

PR #215, opened 2024-01-17
QASM support for singly-controlled ControlledGates.

### LINKED ISSUE #214: Is it possible to get the QASM string representation of a BQSKit circuit?
Hi,

Is it possible to get the QASM string representation of a BQSKit circuit?

mtweiden:
This PR adds support for outputting qasm of singly-controlled "standard" gates. Standard gates are those that have a defined `qasm_name` property.

Resolves #214

### DIFF
--- a/bqskit/ir/gates/composed/controlled.py
+++ b/bqskit/ir/gates/composed/controlled.py
@@ -287,22 +287,32 @@ class ControlledGate(ComposedGate, DifferentiableUnitary):
             ctrl_U = np.kron(self.ctrl, U) + self.ihalf
             self._utry = UnitaryMatrix(ctrl_U, self.radixes)
 
-    def get_qasm(self, params: RealVector, location: CircuitLocation) -> str:
+    @property
+    def qasm_name(self) -> str:
         """
-        Override default `Gate.get_qasm` method.
+        Override default `Gate.qasm_name` method.
 
-        If the target gate is a standard gate, this function will output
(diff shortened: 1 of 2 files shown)


edyounis:
Thanks for being quick with this fix! One couple changes: I would override `Gate.qasm_name` or add an attribute to the class `_qasm_name`. This is because you are not changing how the entire qasm line gets printed.

Looking at the [decoding](https://github.com/BQSKit/bqskit/blob/1ee59f11da206c3b18667c7691aded816016c8ed/bqskit/ir/lang/qasm2/visitor.py#L191), the `cu1`, `cu2`, `cu3`, `cswap`, `c3x`, and `c4x` are all the standard qasm2 identitifers that are handled this way. Does your implementation catch the `cswap`, `c3x`, and `c4x`?

I do like that your implementation handles multiple controls on gates nicely, but I worry this will also create a situation where we can encode something we can decode, e.g. `ControlledGate(XGate(), 5)`. As annoying and limiting as this sounds, I think we should just check for the standard qasm ids, and print a helpful (or annoying but more specific) error in other cases.

Thanks again for putting this up!

mtweiden:
This new version overrides the `Gate.qasm_name` property. There's a check now to ensure that only standard OpenQASM2.0 identifiers (or at least those in `bqskit.ir.lang.qasm2.visitor.py`) can be encoded.

PR #220, opened 2024-01-29
Feature: Cache for runtime workers

mtweiden:
This PR adds the ability to interact with a "runtime cache" for workers. Objects that are not CPickle-able (such as Pytorch modules) can instead be loaded to cache. Doing so ensures that objects are only loaded by each worker a single time. The cache is implemented as a simple Python dictionary.

Use:
In a worker process, items in worker cache can be accessed using the following general outline:
```
from bqskit.runtime import get_runtime

worker_cache = get_runtime().get_cache()

if 'object_name' not in worker_cache:
    object = load_object()
    worker_cache['object_name'] = object
else:
    object = worker_cache['object_name']
```

### DIFF
--- a/bqskit/runtime/__init__.py
+++ b/bqskit/runtime/__init__.py
@@ -166,9 +166,32 @@ class RuntimeHandle(Protocol):
         """
         Retrieve worker's local cache.
 
-        A worker's local cache is implemented as a dictionary. This function can
-        be used to check the presence of data, or to place data directly into, a
-        worker's local cache.
+        In situations where an large or non-easily serializable object is
+        needed during the execution of a custom pass, `get_cache` can be used
+        to load and store data. Objects placed in cache will last from when
+        they are  loaded until the end of a Compiler's Workflow execution.
+        Things placed  in cache are loaded once, then can be reused by
+        subsequent passes.
+
+        For example, say some CustomPassA needs a SomeBigScaryObject. Within
+        its run() method, the object is constructed, then placed into cache.
+        A later pass, CustomPassB, also needs to use SomeBigScaryObject. It
+        checks that the object has been loaded by CustomPassA, then uses it.
+
+        ```
+            # In CustomPassA's .run() definition...
+            unpicklable_object = SomeBigScaryObject(...)
+            worker_cache = get_runtime().get_cache()
+            worker_cache['big_scary_object'] = unpicklable_object
+            unpicklable_object.use(...)
+
+            # In CustomPassB's .run() definition...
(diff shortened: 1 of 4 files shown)


bqskit/runtime/__init__.py
```diff
@@ -162,6 +162,16 @@ def cancel(self, future: RuntimeFuture) -> None:
...
         ...
 
+    def get_cache(self) -> dict[str, Any]:
+        """
+        Retrieve worker's local cache.
+
+        A worker's local cache is implemented as a dictionary. This function can
+        be used to check the presence of data, or to place data directly into, a
+        worker's local cache.
+        """
```
edyounis, line 172: What is a "worker's local cache"? The documentation here should explain how this can be used while abstracting details of how the runtime works. Consider a user reading this may not know (nor should they know) what a worker is. What is important is that this cache is transient. You can place things in here, to reduce reloading, but they may not be in here later. This informs the user that when using this cache, you should write something like:

```python
def load_expensive_io_model_or_something_or_another(file):
    cache = get_runtime().get_cache()
    if some_file_name_key in cache:
        return that
    # load it; store in cache; return it
    ...
```

Not saying you can't or shouldn't put implementation details later in the docstring, but the first points should inform the user how to use this method/function/class/etc. In this case, definitely include somewhere that this cache is local to a specific worker and is never serialized/pickled.

bqskit/runtime/worker.py
```diff
@@ -563,6 +566,16 @@ def cancel(self, future: RuntimeFuture) -> None:
         msgs = [(RuntimeMessage.CANCEL, addr) for addr in addrs]
         self._outgoing.extend(msgs)
 
+    def get_cache(self) -> dict[str, Any]:
+        """
+        Retrieve worker's local cache.
+
+        Returns:
+            (dict[str, Any]): The worker's local cache implemented
+                with a dictionary.
```
edyounis, line 575: It's totally me being pedantic, feel free to ignore this completely, but I did want to mention that you do not need to say it's implemented as a dictionary when the type information is right there.

edyounis (changes requested):
For the most part LGTM, thanks for adding this feature. Please just update the `get_cache` documentation.

bqskit/runtime/__init__.py
mtweiden: Removed reference to workers here. Added example as to how to use the cache.

bqskit/runtime/worker.py
mtweiden: Moved references to worker specific cache here.

PR #234, opened 2024-03-18
Add ECR gate

vincentgierisch:
Add echoed-resonance (ECR) gate

### DIFF
--- a/bqskit/ir/gates/constant/__init__.py
+++ b/bqskit/ir/gates/constant/__init__.py
@@ -14,6 +14,7 @@ from bqskit.ir.gates.constant.cx import CNOTGate
 from bqskit.ir.gates.constant.cx import CXGate
 from bqskit.ir.gates.constant.cy import CYGate
 from bqskit.ir.gates.constant.cz import CZGate
+from bqskit.ir.gates.constant.ecr import ECRGate
 from bqskit.ir.gates.constant.h import HGate
 from bqskit.ir.gates.constant.identity import IdentityGate
 from bqskit.ir.gates.constant.iswap import ISwapGate
@@ -57,6 +58,7 @@ __all__ = [
     'CXGate',
     'CYGate',
     'CZGate',
+    'ECRGate',
     'HGate',
     'IdentityGate',
     'ISwapGate',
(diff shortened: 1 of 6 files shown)


bqskit/ir/gates/constant/__init__.py
```diff
@@ -14,6 +14,7 @@
 from bqskit.ir.gates.constant.cx import CXGate
 from bqskit.ir.gates.constant.cy import CYGate
 from bqskit.ir.gates.constant.cz import CZGate
+from bqskit.ir.gates.constant.ecr import ECRGate
```
edyounis, line 17: Please also add the ECRGate to the docstring in `bqskit/ir/gates/__init__.py`, so that it can be included with documentation.

bqskit/ir/gates/constant/ecr.py
```diff
@@ -0,0 +1,26 @@
+"""This module implements the ECRGate."""
+from bqskit.ir.gates.qubitgate import QubitGate
+from bqskit.ir.gates.constantgate import ConstantGate
+from bqskit.qis.unitary.unitarymatrix import UnitaryMatrix
+import math
+
+class ECRGate(ConstantGate, QubitGate):
```
edyounis, line 7: Please add some documentation that references a good resource for this gate and includes the matrix representation.

edyounis (changes requested):
Awesome! Thanks for taking this upon yourself to add. This is very much a welcome addition, thanks also for getting the unitary endian ordering correct.

Besides the two minor points below, can you add a test that ensures we can read and write this gate properly to qasm. You may need to a GateDef in the [qasm visitor module](https://github.com/BQSKit/bqskit/blob/d52c597e65b5f6b68147754249e7ec55160c5506/bqskit/ir/lang/qasm2/visitor.py#L212).

edyounis:
Awesome! Just one more test to get this approved and merged in. Can you do a encode-decode-in-one test? As in create a circuit with an `ECRGate`, save to qasm with `Circuit.to`, then load that qasm in. I do expect that the resulting circuit will be composed of a CircuitGate with the decomposed gates, but it is good to check there are no crashes and that the input and output unitaries are the same.

PR #236, opened 2024-03-28
Power gate

ermalrrapaj:
Added the power gate class and also included tests.


edyounis (approved):

edyounis (changes requested):
The changes look good. Before we merge this into main, can you fix the pre-commit issues? They don't seem serious, but let me know if you need any help. The tests are also failing, we should take a look at those.

edyounis (approved):

PR #237, opened 2024-03-28
Runtime updates

edyounis:
Improves the runtime by addressing some shortcomings that cause issues in rare situations:
- Splits the single-threaded worker into two threads: one thread does work, while one consumes incoming messages. This prevents the rx buffer from filling up and blocking other parts of the system while the worker is working on a long job.
- Added read receipts to the waiting messages to create a "happens-before" relationship with certain messages. This ensures eventual consistency with idle worker tracking through the system and avoids a previously allowed race condition. Moreover, the manager can now always adjust their idle count to an accurate amount.
- Workers now send messages instantly rather than waiting until their current job has finished or enters an awaited state. This allows for child tasks to start quicker and log messages to appear instantly.

...draft...more to come

### DIFF
--- a/bqskit/runtime/__init__.py
+++ b/bqskit/runtime/__init__.py
@@ -111,7 +111,7 @@ os.environ['MKL_NUM_THREADS'] = '1'
 os.environ['NUMEXPR_NUM_THREADS'] = '1'
 os.environ['VECLIB_MAXIMUM_THREADS'] = '1'
 os.environ['RUST_BACKTRACE'] = '1'
-
+print("SETTING THREADS TO 1")
 
 if TYPE_CHECKING:
     from bqskit.runtime.future import RuntimeFuture
--- a/bqskit/runtime/__init__.py
+++ b/bqskit/runtime/__init__.py
@@ -111,7 +111,7 @@ os.environ['MKL_NUM_THREADS'] = '1'
 os.environ['NUMEXPR_NUM_THREADS'] = '1'
 os.environ['VECLIB_MAXIMUM_THREADS'] = '1'
 os.environ['RUST_BACKTRACE'] = '1'
-print("SETTING THREADS TO 1")
+print('SETTING THREADS TO 1')
 
 if TYPE_CHECKING:
     from bqskit.runtime.future import RuntimeFuture
--- a/bqskit/runtime/base.py
+++ b/bqskit/runtime/base.py
@@ -71,6 +71,7 @@ class RuntimeEmployee:
 def send_outgoing(node: ServerBase) -> None:
     """Outgoing thread forwards messages as they are created."""
     while True:
+        node.logger.debug('Waiting to send outgoing message...')
         outgoing = node.outgoing.get()
 
         if not node.running:
@@ -80,9 +81,16 @@ def send_outgoing(node: ServerBase) -> None:
             # while condition.
             break
 
+        node.logger.debug(f'Sending message {outgoing[1].name}...')
         outgoing[0].send((outgoing[1], outgoing[2]))
         node.logger.debug(f'Sent message {outgoing[1].name}.')
-        node.logger.log(1, f'{outgoing[2]}\n')
+
+        if outgoing[1] == RuntimeMessage.SUBMIT_BATCH:
+            node.logger.log(1, f'{len(outgoing[2])}\n')
+        else:
+            node.logger.log(1, f'{outgoing[2]}\n')
+
+        node.outgoing.task_done()
 
 
 def sigint_handler(signum: int, _: FrameType | None, node: ServerBase) -> None:
@@ -347,6 +355,7 @@ class ServerBase:
         try:
             while self.running:
                 # Wait for messages
+                self.logger.debug('Waiting for messages...')
                 events = self.sel.select()  # Say that 5 times fast
 
                 for key, _ in events:
@@ -368,7 +377,10 @@ class ServerBase:
                         continue
                     log = f'Received message {msg.name} from {direction.name}.'
                     self.logger.debug(log)
-                    self.logger.log(1, f'{payload}\n')
+                    if msg == RuntimeMessage.SUBMIT_BATCH:
+                        self.logger.log(1, f'{len(payload)}\n')
+                    else:
+                        self.logger.log(1, f'{payload}\n')
 
                     # Handle message
                     self.handle_message(msg, direction, conn, payload)
@@ -513,9 +525,10 @@ class ServerBase:
         """Schedule tasks between this node's employees."""
         if len(tasks) == 0:
             return
-
+        self.logger.info(f'Scheduling {len(tasks)} tasks with {self.num_idle_workers} idle workers.')
         assignments = self.assign_tasks(tasks)
 
+        # for e, assignment in sorted(zip(self.employees, assignments), key=lambda x: x[0].num_idle_workers, reverse=True):
         for e, assignment in zip(self.employees, assignments):
             num_tasks = len(assignment)
 
@@ -528,6 +541,7 @@ class ServerBase:
             e.num_idle_workers -= min(num_tasks, e.num_idle_workers)
 
         self.num_idle_workers = sum(e.num_idle_workers for e in self.employees)
+        self.logger.info(f'Finished scheduling {len(tasks)} tasks with now {self.num_idle_workers} idle workers.')
 
     def send_result_down(self, result: RuntimeResult) -> None:
         """Send the `result` to the appropriate employee."""
--- a/bqskit/runtime/base.py
+++ b/bqskit/runtime/base.py
@@ -41,14 +41,20 @@ class RuntimeEmployee:
         conn: Connection,
         total_workers: int,
         process: Process | None = None,
-        num_tasks: int = 0,
     ) -> None:
         """Construct an employee with all resources idle."""
         self.conn: Connection = conn
         self.total_workers = total_workers
         self.process = process
-        self.num_tasks = num_tasks
+        self.num_tasks = 0
         self.num_idle_workers = total_workers
+
+        self.submit_cache: list[tuple[RuntimeAddress, int]] = []
+        """
+        Tracks recently submitted tasks by id and count.
+        This is used to adjust the idle worker count when
+        the employee sends a waiting message.
+        """
 
     def shutdown(self) -> None:
         """Shutdown the employee."""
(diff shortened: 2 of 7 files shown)


bqskit/runtime/__init__.py
```diff
@@ -111,7 +111,7 @@
 os.environ['NUMEXPR_NUM_THREADS'] = '1'
 os.environ['VECLIB_MAXIMUM_THREADS'] = '1'
 os.environ['RUST_BACKTRACE'] = '1'
-
+print('SETTING THREADS TO 1')
```
alonkukl, line 114: Please remove the print

bqskit/runtime/base.py
```diff
@@ -82,7 +104,13 @@ def send_outgoing(node: ServerBase) -> None:
 
         outgoing[0].send((outgoing[1], outgoing[2]))
         node.logger.debug(f'Sent message {outgoing[1].name}.')
-        node.logger.log(1, f'{outgoing[2]}\n')
+
+        if outgoing[1] == RuntimeMessage.SUBMIT_BATCH:
```
alonkukl, line 108: Perhaps also add a sample of the first and last element in the batch?

bqskit/runtime/base.py
```diff
@@ -82,7 +104,13 @@ def send_outgoing(node: ServerBase) -> None:
 
         outgoing[0].send((outgoing[1], outgoing[2]))
         node.logger.debug(f'Sent message {outgoing[1].name}.')
```
alonkukl, line 106: It would be benefitial to also indicate to where the message is sent up\down, and if down to which id (if possible)

bqskit/runtime/base.py
```diff
@@ -368,7 +397,10 @@ def run(self) -> None:
                         continue
                     log = f'Received message {msg.name} from {direction.name}.'
                     self.logger.debug(log)
-                    self.logger.log(1, f'{payload}\n')
+                    if msg == RuntimeMessage.SUBMIT_BATCH:
```
alonkukl, line 400: Same here as above

bqskit/runtime/base.py
```diff
@@ -513,9 +545,9 @@ def schedule_tasks(self, tasks: Sequence[RuntimeTask]) -> None:
         """Schedule tasks between this node's employees."""
         if len(tasks) == 0:
             return
-
         assignments = self.assign_tasks(tasks)
 
+        # for e, assignment in sorted(zip(self.employees, assignments), key=lambda x: x[0].num_idle_workers, reverse=True):
```
alonkukl, line 550: What is this? :)

bqskit/runtime/manager.py
```diff
@@ -126,13 +130,15 @@ def handle_message(
...
             if msg == RuntimeMessage.SUBMIT:
                 rtask = cast(RuntimeTask, payload)
+                self.most_recent_read_submit = rtask.unique_id
                 self.schedule_tasks([rtask])
-                self.update_upstream_idle_workers()
+                # self.update_upstream_idle_workers()
 
             elif msg == RuntimeMessage.SUBMIT_BATCH:
                 rtasks = cast(List[RuntimeTask], payload)
+                self.most_recent_read_submit = rtasks[0].unique_id
```
alonkukl, line 139: should it be -1?

bqskit/runtime/worker.py
```diff
@@ -122,15 +125,56 @@ def deposit_result(self, result: RuntimeResult) -> None:
             self.result[slot_id] = result.result
 
 
+def handle_incoming_comms(worker: Worker) -> None:
+    """Handle all incoming messages."""
+    while True:
+        # Handle incomming communication
```
alonkukl, line 131: Maybe change the comment to say that this is blocking

bqskit/runtime/worker.py
```diff
@@ -122,15 +125,56 @@ def deposit_result(self, result: RuntimeResult) -> None:
             self.result[slot_id] = result.result
 
 
+def handle_incoming_comms(worker: Worker) -> None:
```
alonkukl, line 128: Add more comments on hy we need this function out of the class

alonkukl: Actualy, why is this function not a member if the class?

bqskit/runtime/worker.py
```diff
@@ -122,15 +125,56 @@ def deposit_result(self, result: RuntimeResult) -> None:
...
+            worker.read_receipt_mutex.acquire()
+            tasks = cast(List[RuntimeTask], payload)
+            worker.most_recent_read_submit = tasks[0].unique_id
+            worker._add_task(tasks.pop())  # Submit one task
+            worker._delayed_tasks.extend(tasks)  # Delay rest
+            # Delayed tasks have no context and are stored (more-or-less)
+            # as a function pointer together with the arguments.
+            # When it gets started, it consumes much more memory,
+            # so we delay the task start until necessary (at no cost)
+            worker.read_receipt_mutex.release()
```
alonkukl, line 156: Why do you release the murex only at the end?

bqskit/runtime/worker.py
```diff
@@ -178,16 +222,18 @@ def __init__(self, id: int, conn: Connection) -> None:
         self._id = id
         self._conn = conn
 
-        self._outgoing: list[tuple[RuntimeMessage, Any]] = []
-        """Stores outgoing messages to be handled by the event loop."""
+        # self._outgoing: list[tuple[RuntimeMessage, Any]] = []
```
alonkukl, line 225: Remove code

bqskit/runtime/worker.py
```diff
@@ -178,16 +222,18 @@ def __init__(self, id: int, conn: Connection) -> None:
...
+        # """Stores outgoing messages to be handled by the event loop."""
 
         self._tasks: dict[RuntimeAddress, RuntimeTask] = {}
         """Tracks all started, unfinished tasks on this worker."""
 
         self._delayed_tasks: list[RuntimeTask] = []
         """Store all delayed tasks in LIFO order."""
 
-        self._ready_task_ids: WorkerQueue = WorkerQueue()
+        # self._ready_task_ids: WorkerQueue = WorkerQueue()
```
alonkukl, line 235: ?

bqskit/runtime/worker.py
```diff
@@ -218,11 +270,19 @@ def record_factory(*args: Any, **kwargs: Any) -> logging.LogRecord:
...
 
         logging.setLogRecordFactory(record_factory)
 
+        # Start incoming thread
+        self.incomming_thread = Thread(
+            target=handle_incoming_comms,
+            args=(self,),
+        )
+        self.incomming_thread.start()
+        # self.logger.info('Started incoming thread.')
```
alonkukl, line 284: ?

bqskit/runtime/worker.py
```diff
@@ -338,16 +367,24 @@ def _handle_cancel(self, addr: RuntimeAddress) -> None:
...
-                    continue
-                return None
+            if self._ready_task_ids.empty() and len(self._delayed_tasks) > 0:
+                self._add_task(self._delayed_tasks.pop())
+                continue
+
+            self.read_receipt_mutex.acquire()
+            try:
+                addr = self._ready_task_ids.get_nowait()
+            except Empty:
```
alonkukl, line 380: I don't understad whats going on here, should we talk about it?

alonkukl (changes requested):

bqskit/runtime/base.py
```diff
@@ -41,15 +41,22 @@ def __init__(
...
     ) -> None:
         """Construct an employee with all resources idle."""
         self.conn: Connection = conn
         self.total_workers = total_workers
         self.process = process
-        self.num_tasks = num_tasks
+        self.num_tasks = 0
         self.num_idle_workers = total_workers
 
+        self.submit_cache: list[tuple[RuntimeAddress, int]] = []
```
jkalloor3, line 52: Should this be a dictionary instead of a list of tuples?

jkalloor3: Or maybe ordering matters, so an ordered dict?

bqskit/runtime/base.py
jkalloor3: Agreed, or some more information other than len should be logged

bqskit/runtime/base.py
```diff
@@ -41,15 +41,22 @@ def __init__(
         conn: Connection,
         total_workers: int,
         process: Process | None = None,
-        num_tasks: int = 0,
     ) -> None:
         """Construct an employee with all resources idle."""
         self.conn: Connection = conn
         self.total_workers = total_workers
         self.process = process
-        self.num_tasks = num_tasks
```
jkalloor3, line 50: Why do we not set the number of tasks anymore?

bqskit/runtime/manager.py
```diff
@@ -153,20 +159,21 @@ def handle_message(
...
                 result = cast(RuntimeResult, payload)
                 self.handle_result_from_below(result)
 
             elif msg == RuntimeMessage.WAITING:
-                num_idle = cast(int, payload)
-                self.handle_waiting(conn, num_idle)
+                p = cast(tuple[int, Optional[RuntimeAddress]], payload)
+                num_idle, read_receipt = p
+                self.handle_waiting(conn, num_idle, read_receipt)
                 self.update_upstream_idle_workers()
```
jkalloor3, line 177: Do we need line 177 either? Seems like it should be caught by line 227 below

bqskit/runtime/manager.py
```diff
@@ -217,6 +224,7 @@ def send_up_or_schedule_tasks(self, tasks: Sequence[RuntimeTask]) -> None:
         if num_idle != 0:
             self.outgoing.put((self.upstream, RuntimeMessage.UPDATE, num_idle))
             self.schedule_tasks(tasks[:num_idle])
+            self.update_upstream_idle_workers()
```
jkalloor3, line 227: On Submit msgs we do not set num_idle correct? We do not have to update upstream because the upstream now has an accurate number of idle workers bc of read receipts?

bqskit/runtime/base.py
edyounis: I don't think so. This isn't as much a key-value pair as just a pair of information.

bqskit/runtime/base.py
edyounis: RuntimeEmployee objects are constructed only at startup before any tasks are assigned. It was cluttering the spec in the past.

bqskit/runtime/manager.py
edyounis: In this case, the submit was received from below, and we are scheduling it on the manager locally. Therefore, the manager's boss doesn't know about the updated idle count. For example, if we have one job go to the first worker, who spawns 30 other jobs. The manager may decide to schedule that all locally without the server/higher-manager knowing. Therefore, we now have 30 fewer idle workers on this manager and must inform the server. But when we do send the WAITING message upstream, that is tagged with the last submit call we received from the boss, so it can accurately update its idle count for this manager with information that the manager may not have processed yet.

PR #240, opened 2024-04-09
Adding a TreeScan Gate Removal pass to parallelize Scanning Gate

jkalloor3:
For several circuits of medium width 7-20ish qubits, GPU-enhanced QFactor allows us to run instantiation quickly in order to do a large-block ScanningGate removal. For these workflows, there are very few blocks, which limits our parallelism. In order to fix this, this PR adds a tree structure to ScanningGate removal. The TreeScan works as follows:

1. Split the circuit operations into chunks of size tree_depth
2. At every iteration:
a. Look at the next chunk of operations 
b. Generate 2 ^ tree_depth circuits. Each circuit corresponds to every combination of whether or not to include one of the operations in the chunk.
c. Instantiate in parallel all 2^tree_depth circuits
d. Choose the circuit that has the least number of operations and move on to the next chunk of operations.

This optimization is less greedy than the current ScanningGate removal, which we see can offer much better quality circuits than ScanningGate. In very rare occasions, ScanningGate may be able to outperform TreeScan (since it is still greedy), but in general we can expect TreeScan to almost always outperform ScanningGate.


bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,190 @@
...
+from bqskit.utils.typing import is_real_number
+
+_logger = logging.getLogger(__name__)
+
+
+class TreeScanningGateRemovalPass(BasePass):
+    """
+    The ScanningGateRemovalPass class.
+
+    Starting from one side of the circuit, attempt to remove gates one-by-one.
```
edyounis, line 24: Update documentation, most of your PR comment can fit well here

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,190 @@
...
+                successful removal of a gate.
+                (Default: HilbertSchmidtResidualsGenerator())
+
+            instantiate_options (dict[str: Any]): Options passed directly
+                to circuit.instantiate when instantiating circuit
+                templates. (Default: {})
+
+            tree_depth (int): The depth of the tree of potential
+                solutions to instantiate. Note that 2^(tree_depth) - 1
+                circuits will be instantiated in parallel.
```
edyounis, line 59: What is the default? Can you give a more user-friendly example of what a specific number means? Like what if I set this to three, what does it do there? If this is adequately described in the class docstring, then you can go less verbose here. Also should the default be 1, isn't that just scanning?

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,190 @@
...
+
+        self.collection_filter = collection_filter or default_collection_filter
+
+        if not callable(self.collection_filter):
+            raise TypeError(
+                'Expected callable method that maps Operations to booleans for'
+                ' collection_filter, got %s.' % type(self.collection_filter),
+            )
+
+        self.tree_depth = tree_depth
```
edyounis, line 95: Add type check. You can use `bqskit.utils.typing.is_integer`

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,190 @@
...
+            'dist_tol': self.success_threshold,
+            'min_iters': 100,
+            'cost_fn_gen': self.cost,
+        }
+        self.instantiate_options.update(instantiate_options)
+
+    @staticmethod
+    def get_tree_circs(
+        orig_num_cycles: int, circuit_copy: Circuit,
+        cycle_and_ops: list[tuple[int, Operation]],
```
edyounis, line 109: To keep with the rest of the repo's style, we do one parameter per line in multi-line parameters. No need to change, but worth mentioning

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,190 @@
...
+                ' collection_filter, got %s.' % type(self.collection_filter),
+            )
+
+        self.tree_depth = tree_depth
+        self.start_from_left = start_from_left
+        self.success_threshold = success_threshold
+        self.cost = cost
+        self.instantiate_options: dict[str, Any] = {
+            'dist_tol': self.success_threshold,
+            'min_iters': 100,
```
edyounis, line 101: `min_iters` can probably be 10 or even 0

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,190 @@
...
+        return all_circs
+
+    async def run(self, circuit: Circuit, data: PassData) -> None:
+        """Perform the pass's operation, see :class:`BasePass` for more."""
+        instantiate_options = self.instantiate_options.copy()
+        if 'seed' not in instantiate_options:
+            instantiate_options['seed'] = data.seed
+
+        start = 'left' if self.start_from_left else 'right'
+        _logger.debug(f'Starting scanning gate removal on the {start}.')
```
edyounis, line 136: Update log msg

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,190 @@
...
+
+        target = self.get_target(circuit, data)
+        # target = None
```
edyounis, line 139: Remove unnecessary line

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,190 @@
...
+
+        circuit_copy = circuit.copy()
+        reverse_iter = not self.start_from_left
+
+        ops_left = list(circuit.operations_with_cycles(reverse=reverse_iter))
+        print(
+            f'Starting Scan with tree depth {self.tree_depth}'
+            ' on circuit with {len(ops_left)} gates',
```
edyounis, line 147: The second string in the multiline string is not an f-string. You need to prepend `f` on all strings in the multiline.

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,190 @@
...
+        )
+
+        while ops_left:
+            chunk = ops_left[:self.tree_depth]
+            ops_left = ops_left[self.tree_depth:]
+
+            # Circuits of size 2 ** tree_depth - 1,
```
edyounis, line 154: "Circuits of size" is a bit misleading, it might be better to say "2 ** tree_depth - 1 Circuits".

Moreover, this comment would be redundant with a proper docstring for `get_tree_circs`

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,190 @@
...
+            # ranked in order of most to fewest deletions
+            all_circs = TreeScanningGateRemovalPass.get_tree_circs(
+                circuit.num_cycles, circuit_copy, chunk,
+            )
+            # Remove circuit with no gates deleted
+            all_circs = all_circs[:-1]
```
edyounis, line 160: This should be fused into the function. You can create a wrapper function that executes the two lines of code above. Not that important. Your call to do it or not...

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,190 @@
...
+
+            _logger.debug(
+                'Attempting removal of operation of'
+                f' {self.tree_depth} operations.',
```
edyounis, line 164: I'm not sure if this is descriptive of what is occurring or proper grammar. It might be better to say, "Attempting removal of up to X operations."

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,190 @@
...
+            )
+
+            dists = [self.cost(c, target) for c in instantiated_circuits]
+
+            # Pick least count with least dist
+            for i, dist in enumerate(dists):
+                if dist < self.success_threshold:
+                    _logger.debug(
+                        f'Successfully switched to circuit {i}'
+                        ' of {2 ** self.tree_depth}.',
```
edyounis, line 181: From a logging perspective, this is not helpful. We should instead say something like... "Successfully removed X CNOTs Y U3s" or something like that. Plus f-string issue.

Consider a situation where a user is running on Perlmutter for 12 hours and the time-limit hits. You go to the logs to try and gather some amount of information about how it went to see if its worth restarting or if it did well. You look at the logs and you see a bunch of "Successfully switched to circuit X of of {2 ** Y}." That may not be helpful. Instead if I can count how many successfully removed CNOTs, then I can gather that in 12 hours I made Z amount of progress or something. I know this means that you will have to do more than a simple one-line log call to gather this information, but it is worth doing good logging.

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,190 @@
...
+            'cost_fn_gen': self.cost,
+        }
+        self.instantiate_options.update(instantiate_options)
+
+    @staticmethod
+    def get_tree_circs(
+        orig_num_cycles: int, circuit_copy: Circuit,
+        cycle_and_ops: list[tuple[int, Operation]],
+    ) -> list[Circuit]:
+        # Implement recursively for now, if slow then fix
```
edyounis, line 111: - Remove comment, also this isn't recursive
- Add docstring

bqskit/passes/__init__.py
```diff
@@ -329,6 +330,7 @@
     'UpdateDataPass',
     'ToU3Pass',
     'ScanningGateRemovalPass',
+    'TreeScanningGateRemovalPass',
```
edyounis, line 333: Also add the to the documentation docstring in this file

edyounis (changes requested):
Good PR. Minor changes requested

bqskit/passes/__init__.py
jkalloor3: Does that just mean adding the TreeScanningGateRemovalPass where I did in my latest commit?

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,226 @@
+"""This module implements the ScanningGateRemovalPass."""
```
edyounis, line 1: Update Docstring

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,226 @@
...
+from bqskit.runtime import get_runtime
+from bqskit.utils.typing import is_integer
+from bqskit.utils.typing import is_real_number
+
+_logger = logging.getLogger(__name__)
+
+
+class TreeScanningGateRemovalPass(BasePass):
+    """
+    The ScanningGateRemovalPass class.
```
edyounis, line 23: Update Docstring

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,226 @@
...
+
+    Starting from one side of the circuit, run the following:
+
+    Split the circuit operations into chunks of size tree_depth
```
edyounis, line 27: When writing documentation, it's good practice to wrap variable names in backticks. Like tree_depth should be `tree_depth`

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,226 @@
...
+    At every iteration:
+    a. Look at the next chunk of operations
+    b. Generate 2 ^ tree_depth circuits. Each circuit corresponds to every
+    combination of whether or not to include one of the operations in the chunk.
+    c. Instantiate in parallel all 2^tree_depth circuits
+    d. Choose the circuit that has the least number of operations and move
+    on to the next chunk of operations.
+
+    This optimization is less greedy than the current ScanningGate removal,
```
edyounis, line 36: For better documentation, you may want to link ScanningGateRemoval using the proper syntax:

```raw
:class:`~bqskit.passes.processing.ScanningGateRemovalPass`
```
We can also double check the documentation before merging by clicking on the PR readthedocs test.

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,226 @@
...
+    which we see can offermuch better quality circuits than ScanningGate.
```
edyounis, line 37: offer much

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,226 @@
...
+    In very rare occasions, ScanningGate may be able to outperform
```
edyounis, line 38: I don't understand why ScanningGate would be able to outperform TreeScan. Isn't every candidate considered during ScanningGate considered during TreeScan? I know there is always variance in instantiation so that anything can happen for sure, but is there something I am missing?

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,226 @@
...
+        self,
+        start_from_left: bool = True,
+        success_threshold: float = 1e-8,
+        cost: CostFunctionGenerator = HilbertSchmidtResidualsGenerator(),
+        instantiate_options: dict[str, Any] = {},
+        tree_depth: int = 1,
+        collection_filter: Callable[[Operation], bool] | None = None,
+    ) -> None:
+        """
+        Construct a ScanningGateRemovalPass.
```
edyounis, line 53: Update Docstring

bqskit/passes/processing/treescan.py
```diff
@@ -0,0 +1,226 @@
...
+    def get_tree_circs(
+        orig_num_cycles: int,
+        circuit_copy: Circuit,
+        cycle_and_ops: list[tuple[int, Operation]],
+    ) -> list[Circuit]:
+        '''
+        Given a circuit, create 2^(tree_depth) - 1 circuits that remove up
+        to tree_depth operations. The circuits are sorted by the number of
+        operations removed.
+        '''
```
edyounis, line 140: Function and method docstrings should start with a single sentence summary, then more if needed after a blank line, followed by argument descriptions. More sections can follow if necessary, e.g., Returns, Examples, See Also, etc. I know this is just a helper method now, but I can foresee this method being reused in future projects. That is why I am being pedantic here. We should make it as readable as possible.

bqskit/passes/processing/treescan.py
edyounis: Log should indicate this is a tree scan pass.

edyounis (changes requested):
You did not address a few comments from the previous review. I am not a fan of how GitHub collapses the comments, so you probably missed them. You might need to expand the ones in the middle. Let me know if you need help here.

Otherwise, it looks much better. All of my comments this time around were focused on the documentation. We should make sure the code you write is properly documented; it will ensure others can use, maintain, and expand your code. Once you push up a commit, you can check how the documentation looks by clicking on the readthedocs CI check.

bqskit/passes/processing/treescan.py
jkalloor3: For example, if there are 20 gates and TreeScan looks at 5 at a time. Let's say that the options are that if you remove the first gate, you can also remove the last 10 gates in the circuit. Otherwise, if you don't remove the first gate, you can remove the next 4 gates in the circuit (but this removes your ability to remove the last 10 gates in the circuit). TreeScan intially would pick the 4 gate option over the single gate option, and then miss out removing the last 10 gates. I don't know if this is even possible in practice, and I haven't seen a single case where this does happen, but AFAIK it could be a possibility?

edyounis: I see. Yeah, that makes sense. Thanks for explaining.

PR #244, opened 2024-04-30
SqrtTGate

mtweiden:
The square root T gate has potential applications for fault tolerant compilation. This PR adds it to BQSKit's library of constant gates.


edyounis (dismissed):
I like the qasm name 'st' better than 'sqrtt' to keep with conventions ('sx'), but hey that's entirely preference. LGTM, thanks for adding this.

PR #249, opened 2024-06-03
add get_inverse features

Youngcius:
1. Add get_inverse and get_inverse_params features for all single-qubit and two-qubit constant and parameterized gates
2. Add necessary gate definitions and corresponding OpenQASM definitions
3. Remedy a format-string line in "passes/mapping/placement/trivial.py" that leads to testing errors

### DIFF
--- a/tests/ir/test_gate.py
+++ b/tests/ir/test_gate.py
@@ -137,3 +137,37 @@ class TestBasicGate:
         utry = gate.get_unitary(params)
         pickled_utry = pickle.loads(pickle.dumps(gate)).get_unitary(params)
         assert utry == pickled_utry
+
+    def test_get_inverse(self, gate: Gate) -> None:
+        from bqskit.ir.gates.constantgate import ConstantGate
+        from bqskit.qis.unitary.differentiable import DifferentiableUnitary
+        from bqskit.ir.gates.qubitgate import QubitGate
+        from bqskit.ir.operation import Operation
+        if isinstance(gate, QubitGate):
+            if isinstance(gate, ConstantGate):
+                assert hasattr(gate, 'get_inverse')
+                inv_gate = gate.get_inverse()
+                assert inv_gate._qasm_name, \
+                    'All inverses should have an effective _qasm_name'
+                iden = np.identity(gate.dim)
+                inv_gate = gate.get_inverse()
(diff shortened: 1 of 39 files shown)


tests/ir/test_gate.py
```diff
@@ -137,3 +137,37 @@ def test_pickle(self, gate: Gate) -> None:
...
+                assert hasattr(gate, 'get_inverse')
+                inv_gate = gate.get_inverse()
+                assert inv_gate._qasm_name, \
+                    'All inverses should have an effective _qasm_name'
+                iden = np.identity(gate.dim)
+                inv_gate = gate.get_inverse()
+                supposed_to_be_iden = (
+                    inv_gate.get_unitary() @ gate.get_unitary()
+                )
+                dist = supposed_to_be_iden.get_distance_from(iden, 1)
```
edyounis, line 157: Is `get_distance_from` the correct way to test here? This will ignore global phase, I have to think we should allow this. If you make the test more strict and just check the norm to identity, for example, do you have any issues?

Also, we should test multiplication both ways, A@B=I and B@A=I.

tests/ir/test_gate.py
```diff
@@ -137,3 +137,37 @@ def test_pickle(self, gate: Gate) -> None:
...
+                    'All inverses should have an effective _qasm_name'
+                op = Operation(
+                    gate, list(range(gate.num_qudits)),
+                    np.random.rand(gate.num_params),
+                )
+                inv_op = op.get_inverse()
+                iden = np.identity(gate.dim)
+                supposed_to_be_iden = (inv_op.get_unitary() @ op.get_unitary())
+                dist = supposed_to_be_iden.get_distance_from(iden, 1)
+                assert dist < 1e-10
```
edyounis, line 173: Can we also do another test for the qasm definitions you added? I think we wrote that test before we even had qasm fully working, so it just checks that its a string. We should ensure that a circuit containing these gates (one at a time) can be saved to and loaded from qasm, while also maintaining the same unitary definition before and after.

edyounis (changes requested):
Awesome work! Thanks for doing this!

Address the comments, then I will do a final check on all the math, and we will be good to go!

Let me know if you have any concerns or questions.

Issue #224, opened 2024-02-27
Subgraph Isomorphism Checks


edyounis:
We can add subgraph isomorphism passes that try to match logical connectivity to a physical topology without performing any swap operations. We can also add these passes to specific spots in the standard workflow to avoid placement + layout + routing.

The main step here is to [develop a compiler pass](https://bqskit.readthedocs.io/en/latest/guides/custompass.html) that tries to find a [subgraph isomorphism](https://en.wikipedia.org/wiki/Subgraph_isomorphism_problem) between a Circuit's logical connectivity and a MachineModel's physical topology.

To get the logical topology, you can call [`Circuit.coupling_graph`](https://github.com/BQSKit/bqskit/blob/f0c681faadaf84766a474364068298527356cd47/bqskit/ir/circuit.py#L234). This gives you a graph where nodes are the qudits in the circuit and edges exist between qudits that have a gate between them in the circuit.

To get the physical topology of the target MachineModel during compilation, you can use the PassData object in a pass's run function: `data.model.coupling_graph`. This gives you a graph where nodes are physical qudits in a QPU, and edges exist between qudits that can interact.

Both of these return [`CouplingGraph`](https://github.com/BQSKit/bqskit/blob/f0c681faadaf84766a474364068298527356cd47/bqskit/qis/graph.py#L31) objects, which has many helpful methods.

The goal is **to attempt to** find a map between logical and physical topology without needing any further routing. This will not always be possible. There are many algorithms that will solve this problem, but we should consider that our problem size may vary greatly and that we should have a time-out of some sort before we fall back to traditional layout and routing strategies already implemented in BQSKit.

If we have found a mapping, we can set `data.initial_mapping = data.final_mapping = our_answer` and skip layout and routing steps in the BQSKit workflows. Integrating this into BQSKit's workflows can be made into a separate issue and doesn't need to be addressed here.

SoshunNaito:
Hi @edyounis, I am very interested in this topic.

The easiest way to realize that idea is to use `networkx.algorithms.isomorphism.GraphMatcher` class, which supports the `subgraph_is_monomorphic` function.
I think we should use monomorphism instead of isomorphism because extra physical edges do not affect the compilation result at all. For example, `G1=[a-b c-d]` can be embedded into `G2=[v w-x-y-z]`, but this mapping is not isomorphic because no edge in G1 is mapped to `x-y`.

[...]

PR #250, opened 2024-06-04
add StaticPlacementPass

SoshunNaito:
This PR adds a new pass named `StaticPlacementPass`, which attempts to find an embedding of logical qubits into the physical coupling graph so that no SWAPs are needed.
Also, it supports a `timeout_sec` parameter to limit the search time.
Please note that additional packages such as `timeout_decorator` and `networkx` are required to run this pass.

Original issue: https://github.com/bqskit/bqskit/issues/224

### DIFF
--- a/bqskit/passes/__init__.py
+++ b/bqskit/passes/__init__.py
@@ -234,6 +234,7 @@ from bqskit.passes.mapping.embed import EmbedAllPermutationsPass
 from bqskit.passes.mapping.layout.pam import PAMLayoutPass
 from bqskit.passes.mapping.layout.sabre import GeneralizedSabreLayoutPass
 from bqskit.passes.mapping.placement.greedy import GreedyPlacementPass
+from bqskit.passes.mapping.placement.static import StaticPlacementPass
 from bqskit.passes.mapping.placement.trivial import TrivialPlacementPass
 from bqskit.passes.mapping.routing.pam import PAMRoutingPass
 from bqskit.passes.mapping.routing.sabre import GeneralizedSabreRoutingPass
@@ -364,6 +365,7 @@ __all__ = [
     'GeneralizedSabreLayoutPass',
     'GreedyPlacementPass',
     'TrivialPlacementPass',
+    'StaticPlacementPass',
     'GeneralizedSabreRoutingPass',
     'SetModelPass',
     'U3Decomposition',
--- a/bqskit/passes/mapping/__init__.py
+++ b/bqskit/passes/mapping/__init__.py
@@ -6,6 +6,7 @@ from bqskit.passes.mapping.embed import EmbedAllPermutationsPass
 from bqskit.passes.mapping.layout.pam import PAMLayoutPass
 from bqskit.passes.mapping.layout.sabre import GeneralizedSabreLayoutPass
 from bqskit.passes.mapping.placement.greedy import GreedyPlacementPass
+from bqskit.passes.mapping.placement.static import StaticPlacementPass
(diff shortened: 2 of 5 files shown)


edyounis:
This is definitely on the right track. Thanks for taking this issue up and putting a PR.

I like the timeout feature, but we should have this interface with the predicate system to simplify workflows containing this. By this I mean, turning this `BasePass` actually into a `PassPredicate`, or have it work with a predicate. This should also remove the RuntimeErrors. We can then have a workflow that looks like:

```python
workflow = [
    StaticPlacementPass(...)
    IfThenElsePass(
        FoundPlacement(),
        NOOPPass(),
        normal_mapping_workflow,
]
```

I am also very hesitant to add a heavy dependency (networkx) for just this one algorithm. Can we implement the algorithm internally?

SoshunNaito:
> By this I mean, turning this `BasePass` actually into a `PassPredicate`, or have it work with a predicate. This should also remove the RuntimeErrors.

Great idea! I have removed RuntimeErrors for the first step.
In this version, the `StaticPlacementPass` can be used as follows:

```
workflow = [
    SetModelPass(model),
    StaticPlacementPass(),
    IfThenElsePass(
        PhysicalPredicate(),
        [LogPass("Static Placement Found")],
        [LogPass("Greedy Placement Required"), GreedyPlacementPass()],
    ),
    ApplyPlacement(),
]
```


> I am also very hesitant to add a heavy dependency (networkx) for just this one algorithm. Can we implement the algorithm internally?

Yes, of course! We can use an exhaustive search for monomorphism by checking all possible embeddings from logical to physical qudits.
I hope implementing the algorithm internally will remove both `networkx` and `timeout_decorator` dependencies.

SoshunNaito:
Here I have added a naive implementation of the `find_monomorphic_subgraph` method.
Since it requires approximately $(N_p)^{N_l}$ iterations (where $N_p$ and $N_l$ represent the number of physical and logical qudits), we need to further optimize it by pruning the search space.

SoshunNaito:
I have added a pruning feature to reduce the search space.
While placing logical qudits one by one, it skips physical qudits based on the following conditions:
1. The physical qudit must not be occupied by another logical qudit.
2. The degree of the physical qudit must not be less than the degree of the logical qudit.
3. The physical qudit must be connected to all the previous logical qudits.

Using this feature, the search space is reduced from $(N_p)^{N_l}$ to $|S|^{N_l}$, where $|S|$ is the average number of candidates for each step.

SoshunNaito:
Hi @edyounis, how does my code look to you? I would appreciate your comments, feedbacks, or suggestions for further improvement. Thanks!

edyounis:
Looks good, can we add some tests? How large can we push this within a 1- or 10-second time frame?

SoshunNaito:
Sure, of course! I have added `test_static.py` and measured how long it takes for graph embedding.
I used circuits with circular connectivity and physical devices with square grid topology, so it is clear that the circuits can be embedded iff the number of qudits is even.

These two plots below show the elapsed time for each condition `(grid_size, logical_qudits)`.
To answer your question, we can test all possible placements for up to (5x5, 20) in 10 seconds and (5x5, 15) in 1 second.
When there is a valid placement, the embedding runs in less than 1 second even for (7x7, 40).
In addition, when the circuit contains nodes with high degrees, the embedding ends in a short time owing to the pruning feature.
valid_embeddings
invalid_embeddings

edyounis:
Pre-commit and tests are failing, can you get those fixed. The code does look good

SoshunNaito:
@edyounis The code seems to be fixed now. Could you check it?

PR #254, opened 2024-06-28
PauliZGate

mtweiden:
New features:
- A new `PauliZMatrices` class allows for diagonal Hermitian matrices to be easily specified. Using the `PauliMatrices` class results in unwanted overparameterization. The API used is very similar to that for `PauliMatrices`.
- A new `PauliZGate` lets n-qubit diagonal unitaries be specified in a Circuit.
- Support synthesizing diagonal unitaries into quantum circuits [1]. This is facilitated through the `WalshDiagonalSynthesisPass`. Outputs are always use `CNOTGate`s and `RZGate`s.

[1] https://arxiv.org/abs/1306.3991
[2] https://arxiv.org/abs/2109.03371


bqskit/qis/pauliz.py
```diff
@@ -0,0 +1,251 @@
+"""This module implements the PauliMatrices class."""
```
edyounis, line 1: Update documentation

bqskit/qis/pauliz.py
```diff
@@ -0,0 +1,251 @@
...
+if TYPE_CHECKING:
+    from bqskit.qis.unitary.unitary import RealVector
+
+
+class PauliZMatrices(Sequence[npt.NDArray[np.complex128]]):
+    """
+    The group of Pauli Z matrices.
+
+    A PauliZMatrices object represents the entire of set of Pauli Z matrices for
+    some number of qubits.
```
edyounis, line 27: This is not immediately clear what this is. Maybe give an example for 1, and say 3 qubits?

Additionally, if there are relevant references, list them here.

bqskit/qis/pauliz.py
```diff
@@ -0,0 +1,251 @@
...
+        else:
+            self.paulizs = []
+            matrices = it.product(
+                PauliZMatrices(
+                    num_qudits - 1,
+                ),
+                PauliZMatrices(1),
+            )
+            for pauliz_n_1, pauliz_1 in matrices:
+                self.paulizs.append(np.kron(pauliz_n_1, pauliz_1))
```
edyounis, line 87: Should we cache this calculation? I remember we did some caching for the Pauli matrices class.

bqskit/qis/pauliz.py
```diff
@@ -0,0 +1,251 @@
...
+
+        if any(q < 0 or q >= self.num_qudits for q in q_set):
+            raise ValueError('Qubit indices must be in [0, n).')
+
+        if len(q_set) != len(set(q_set)):
+            raise ValueError('Qubit indices cannot have duplicates.')
+
+        # Nth Order Pauli Z Matrices can be thought of base 2 number
+        # I = 0, Z = 1
+        # IZZ = 1 * 2^2 + 1 * 2^1 + 0 * 4^0 = 6 (base 10)
```
edyounis, line 153: 4^0 -> 2^0

bqskit/qis/pauliz.py
```diff
@@ -0,0 +1,251 @@
...
+
+        Args:
+            alpha (RealVector): The pauli coefficients.
+
+        Returns:
+            np.ndarray: Sum of element-wise multiplication of `alpha`
+            and `self.paulis`.
+
+        Raises:
+            ValueError: If `alpha` and `self.paulis` are incompatible.
```
edyounis, line 178: self.paulizs

bqskit/qis/pauliz.py
```diff
@@ -0,0 +1,251 @@
...
+            pauliz_n_qubit.append(self.paulizs[idx])
+
+        return pauliz_n_qubit
+
+    def dot_product(self, alpha: RealVector) -> npt.NDArray[np.complex128]:
+        """
+        Computes the standard dot product of `alpha` with the paulis.
+
+        Args:
+            alpha (RealVector): The pauli coefficients.
```
edyounis, line 171: pauli z coefficients

bqskit/qis/pauliz.py
```diff
@@ -0,0 +1,251 @@
...
+
+    @staticmethod
+    def from_string(
+        pauliz_string: str,
+    ) -> npt.NDArray[np.complex128] | list[npt.NDArray[np.complex128]]:
+        """
+        Construct Pauli Z matrices from a string description.
+
+        Args:
+            pauli_string (str): A string that describes the desired matrices.
```
edyounis, line 202: pauliz_string

bqskit/qis/pauliz.py
```diff
@@ -0,0 +1,251 @@
...
+                This is a comma-seperated list of pauli strings.
```
edyounis, line 203: pauli z

bqskit/passes/synthesis/diagonal.py
```diff
@@ -0,0 +1,108 @@
...
+
+    async def synthesize(
+        self,
+        utry: UnitaryMatrix | StateVector | StateSystem,
+        data: PassData,
+    ) -> Circuit:
+        """Synthesize `utry`, see :class:`SynthesisPass` for more."""
+        if not isinstance(utry, UnitaryMatrix):
+            m = 'DiagonalSynthesisPass can only synthesize diagonal, '
+            m += f'UnitaryMatrixs, got {type(utry)}.'
```
edyounis, line 79: The plural UnitaryMatrixs is bothering me. A nit pick for sure. I know they are object names and as such you just want to add an s to the end of it, but can we surround it with backticks or quotes? "`UnitaryMatrix`s"

Your call

bqskit/passes/synthesis/diagonal.py
```diff
@@ -0,0 +1,108 @@
...
+            raise TypeError(m)
+
+        if not utry.is_qubit_only():
+            m = 'DiagonalSynthesisPass can only synthesize diagonal '
+            m += 'UnitaryMatrixs with qubits, got higher radix than 2.'
```
edyounis, line 84: For future note: can probably generalize the pauliz to qudits and use QFAST. QFAST is great at diagonal synthesis with the pauli z group.

bqskit/passes/synthesis/diagonal.py
```diff
@@ -0,0 +1,108 @@
...
+from bqskit.qis.unitary import UnitaryMatrix
+from bqskit.utils.math import pauliz_expansion
+from bqskit.utils.math import unitary_log_no_i
+
+
+_logger = logging.getLogger(__name__)
+
+
+class DiagonalSynthesisPass(SynthesisPass):
+    """A pass that synthesizes diagonal unitaries."""
```
edyounis, line 24: The class name sounds like a base class. If it performs a specific algorithm, can you adjust the name to reflect that? Additionally, can you include a reference?

bqskit/passes/control/predicates/diagonal.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class DiagonalPredicate(PassPredicate):
+    """
+    The DiagonalPredicate class.
+
+    The DiagonalPredicate class returns True if the circuit's unitary can be
+    approximately inverted by a diagonal unitary. A unitary is approximately
+    inverted when the Hilbert-Schmidt distance to the identity is less than some
+    threshold.
```
edyounis, line 21: Might want to put a "See Also" section in the documentation.

bqskit/ir/gates/parameterized/pauliz.py
```diff
@@ -0,0 +1,103 @@
...
+        if building_docs():
+            self.sigmav: npt.NDArray[Any] = np.array([])
+        else:
+            self.sigmav = (-1j / 2) * self.paulis.numpy
+
+    def get_unitary(self, params: RealVector = []) -> UnitaryMatrix:
+        """Return the unitary for this gate, see :class:`Unitary` for more."""
+        self.check_parameters(params)
+        H = dot_product(params, self.sigmav)
+        eiH = sp.linalg.expm(H)
```
edyounis, line 65: If performance is important for your application, there should be a faster way to calculate `eiH` since `H` is diagonal. I would expect the same to be true for the gradient calculation but I would have to dig a bit deeper there. If it is not important, can you put a "TODO" marker so we can come back to this later.

bqskit/utils/math.py
```diff
@@ -257,3 +288,24 @@ def canonical_unitary(
     correction_phase = 0 - std_phase
     std_correction = np.exp(1j * correction_phase)
     return std_correction * special_unitary
+
+
+def diagonal_distance(unitary: npt.NDArray[np.complex128]) -> float:
+    """
+    Compute how diagonal a unitary is.
```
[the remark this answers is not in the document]
mtweiden: There will be soon!

bqskit/utils/math.py
```diff
@@ -168,6 +169,36 @@ def pauli_expansion(H: npt.NDArray[np.complex128]) -> npt.NDArray[np.float64]:
     return np.array(X)
 
 
+def pauliz_expansion(H: npt.NDArray[np.complex128]) -> npt.NDArray[np.float64]:
+    """
+    Computes a Pauli Z expansion of the diagonal hermitian matrix H.
+
+    Args:
+        H (np.ndarray): The diagonal hermitian matrix to expand.
```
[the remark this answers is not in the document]
mtweiden: Mostly to keep in line with how the `pauli_expansion` function is called, this matrix has shape (N, N). I've added this as clarification.

bqskit/qis/pauliz.py
mtweiden: Can you point me towards this? I don't see it in the Pauli matrices class definition.

bqskit/ir/gates/parameterized/pauliz.py
mtweiden: I've changed the computation for `get_unitary` but have left a TODO for the gradient part.

bqskit/passes/synthesis/diagonal.py
mtweiden: I've changed this to a base class and added a link to the paper which describes this process. There isn't really a common name used to describe this so I've left it as is. For future reference: is this not considered a `SynthesisPass` because no instantiation is needed?

mtweiden: I'm actually changed this back to a `SynthesisPass`, because the use case is `UnitaryMatrix` to `Circuit`, as opposed to `Circuit` to `Circuit`.

bqskit/qis/pauliz.py
edyounis: nvm, we removed it a long time ago. I forget why, but there must have been a reason

bqskit/passes/synthesis/diagonal.py
edyounis: I meant to say that the class name should reflect the algorithm it performs, not the outcome. Yes, this class does diagonal synthesis, but how does it do diagonal synthesis? If we make two algorithms that perform diagonal synthesis, what would we name the other one? To clarify, I am just asking you to change the name of this class, not create a base class.

mtweiden: I see. I've renamed the pass to `WalshDiagonalSynthesisPass`. I've kept the file name as `discrete.py`, but can change that too if you think something like `walshdiagonal.py` is more appropriate.

edyounis (approved):

PR #257, opened 2024-07-11
Surround filter

edyounis:
New features:
- `Circuit.surround` now can take a scoring function and a filter
  - The filter will prune the search space if provided
  - The scoring function can be used to change the scoring mechanism from the default gate count
  - The method is also now faster and more robust as it has been updated with the DAG methods
  - The starting point can now also be a region
- `Circuit.next` and `Circuit.prev` can now also take a `CircuitRegion` and return the next or prev points of the entire region
- Implemented `CouplingGraph.is_linear`

Bug Fixes:
- Fixed issues in circuit grid iteration with `exclude=True`
- Fixed `Circuit.check_region` missing bad regions where a gate would be half-in-half-out

### DIFF
--- a/bqskit/ir/circuit.py
+++ b/bqskit/ir/circuit.py
@@ -2286,28 +2286,68 @@ class Circuit(DifferentiableUnitary, StateVectorMap, Collection[Operation]):
             # Expand node
             for point in self.next(node).union(self.prev(node)):
                 # Create new region by adding the gate at this point
-                new_region_bldr = {k: v for k, v in node.items()}
+                region_bldr = {k: v for k, v in node.items()}
                 op = self[point]
+                valid_region = True
                 for qudit in op.location:
-                    if qudit not in new_region_bldr:
-                        new_region_bldr[qudit] = CycleInterval(
-                            point[0], point[0],
+                    if qudit not in region_bldr:
+                        region_bldr[qudit] = CycleInterval(point[0], point[0])
+
+                    elif point[0] < region_bldr[qudit][0]:
+                        # Check for gates in the middle not in region
+                        if any(
+                            not self.is_point_idle((i, qudit))
+                            for i in range(point[0] + 1, region_bldr[qudit][0])
+                        ):
+                            valid_region = False
+                            break
+
+                        # Absorb Single-qudit gates
+                        index = point[0]
+                        while index > 0:
+                            if not self.is_point_idle((index - 1, qudit)):
+                                prev_op = self[index - 1, qudit]
+                                if len(prev_op.location) != 1:
+                                    break
+                            index -= 1
+
+                        region_bldr[qudit] = CycleInterval(
+                            index,
+                            region_bldr[qudit][1],
                         )
-                    else:
-                        new_region_bldr[qudit] = CycleInterval(
-                            min(new_region_bldr[qudit][0], point[0]),
-                            max(new_region_bldr[qudit][1], point[0]),
+
+                    elif point[0] > region_bldr[qudit][1]:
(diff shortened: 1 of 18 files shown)


bqskit/ir/circuit.py
```diff
@@ -2143,193 +2221,145 @@ def surround(
...
             )
 
-        if bounding_region is not None:
-            bounding_region = CircuitRegion(bounding_region)
+        if filter is not None and not callable(filter):
+            raise TypeError(f'Expected callable filter, got {type(filter)}.')
 
-        point = self.normalize_point(point)
+        def default_scoring_fn(region: CircuitRegion) -> float:
+            return float(sum(op.num_qudits for op in self[region]))
```
mtweiden, line 2228: Maybe this isn't super important as it's just a default, but this means single- and multi- qudit operations are very similarly important.

tests/ir/circuit/test_region_methods.py
```diff
@@ -338,3 +339,59 @@ def test_with_fold(self, r6_qudit_circuit: Circuit) -> None:
...
+
+        # whole wall of even
+        circuit.append_gate(CNOTGate(), [0, 1])
+        circuit.append_gate(CNOTGate(), [2, 3])
+        circuit.append_gate(CNOTGate(), [4, 5])
+
+        region = circuit.surround((1, 3), 4)
+        assert region.location == CircuitLocation([2, 3, 4, 5])
+
+    def test_surround_filter_hard(self) -> None:
```
mtweiden, line 361: What are some other use cases for filtering? It seems like maybe a MachineModel/topology or gateset check would be a nice extra test to have.

bqskit/ir/circuit.py
```diff
@@ -2143,193 +2221,145 @@ def surround(
...
+                        ):
+                            valid_region = False
+                            break
 
-                    # Gates already accounted for stop the half_wire
-                    if (cycle_index, op) in absorbed_gates:
-                        break
+                        # Absorb Single-qudit gates
+                        index = point[0]
+                        while index < self.num_cycles - 1:
```
mtweiden, line 2330: Maybe the parts that do "Absorb Single-qudit gates" and "Check for gates in the middle not in region" should be sub-functions on there own.

bqskit/ir/circuit.py
edyounis: No, this is an important point. I thought going by the number of qudits in a gate would be enough, but your comment made me realize this needs more. For example, a region with two single-qubit gates is equivalent to one with a two-qubit gate. How do you think we should solve this? Multiply by a factor, e.g., x100?

bqskit/ir/circuit.py
edyounis: Yeah, using the flag "valid_region" gave me that same thought, but I can't see a way that makes it any simpler. How are you thinking to do this?

mtweiden: It might actually be most clear as is, since the searches over cycles go in opposite directions. Does the `valid_region = False` on line 2327 ever run in practice? Otherwise LGTM

edyounis: Yeah, that flag is incredibly important. It prevents us from needing to do full, expensive checks all the time. Think of a diamond of cnots (4 qubits, 1->2 ; 0->1, 2->3 ; 1->2). The search can find a U-shape region that goes around one of the cnots in the middle layer. We need to mark this as invalid and prune. If that shape belongs to the largest (goal) region, then it's fine. It's provable that any valid region has a path in search space of smaller valid regions leading up to it.

PR #258, opened 2024-07-14
Constrained minimizer

mtweiden:
This pull request adds support for constrained numerical optimization in numerical instantiation.

### DIFF
--- a/bqskit/ir/opt/constrained.py
+++ b/bqskit/ir/opt/constrained.py
@@ -1,0 +1,48 @@
+"""This module implements the Minimizer base class."""
+from __future__ import annotations
+
+import abc
+from typing import TYPE_CHECKING
+
+import numpy as np
+import numpy.typing as npt
+
+from bqskit.qis.unitary.unitary import RealVector
+
+if TYPE_CHECKING:
+    from bqskit.ir.opt.cost.function import CostFunction
+
+
+class ConstrainedMinimizer(abc.ABC):
+    """
+    The Minimizer class.
+
+    An minimizer finds the parameters for a circuit template that minimizes some
+    CostFunction.
+    """
+
+    @abc.abstractmethod
+    def constrained_minimize(
+        self,
+        cost: CostFunction,
+        cstr: CostFunction,
+        x0: RealVector,
+    ) -> npt.NDArray[np.float64]:
+        """
+        Minimize `cost` starting from initial point `x0` while obeying `cstr`.
+
+        Args:
+            cost (CostFunction): The CostFunction to minimize.
+
+            cstr (CostFunction): The CostFunction used to constrain solutions.
+
+            x0 (np.ndarray): The initial point.
+
+        Returns:
+            (np.ndarray): The inputs that best minimizes the cost while obeying
+                the constraints.
+
+        Notes:
+            This function should be side-effect free. This is because many
+            calls may be running in parallel.
+        """


bqskit/ir/opt/constrained.py
```diff
@@ -0,0 +1,48 @@
...
+
+    An minimizer finds the parameters for a circuit template that minimizes some
+    CostFunction.
+    """
+
+    @abc.abstractmethod
+    def constrained_minimize(
+        self,
+        cost: CostFunction,
+        cstr: CostFunction,
```
alonkukl, line 28: What is the difference between the constraint to the cost? Should we minimize both of them? Maybe the constraint should be a Boolean function meaning that it returns true or false?

bqskit/ir/opt/constrained.py
```diff
@@ -0,0 +1,48 @@
...
+if TYPE_CHECKING:
+    from bqskit.ir.opt.cost.function import CostFunction
+
+
+class ConstrainedMinimizer(abc.ABC):
+    """
+    The Minimizer class.
+
+    An minimizer finds the parameters for a circuit template that minimizes some
+    CostFunction.
```
edyounis, line 21: The documentation should be updated to reflect the new base class.

edyounis (changes requested):

bqskit/ir/opt/constrained.py
mtweiden: The `cstr` cost function will almost always be a unitary distance/fidelity based cost, while the `cost` cost function quantifies some desirable quality of the output circuit to optimize (i.e. it will not be distance based). 

The choice to minimize them both comes from the general form of [constrained optimization problems](https://en.wikipedia.org/wiki/Constrained_optimization). It still makes sense to use a scalar valued cost for the constraint because it's likely we would want to know how far off from being constrained a solution is (or if we want to shift some success threshold). 

I've added some details in the docstring about what the constraint should look like. The cost function is left more open ended.

alonkukl: I'm sorry, I'm still confused.

This abstract function tries to minimize the cost, the constraint, or both?
I assume both, but then why would one want two different cost functions to minimize? do they have different weights? why not combine them into a single cost function?

mtweiden: I'd like to preface by saying that these design choices are based on how off the shelf constrained optimizers work. The [wikipedia page on the KKT conditions](https://en.wikipedia.org/wiki/Karush%E2%80%93Kuhn%E2%80%93Tucker_conditions) and the [documentation for scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.NonlinearConstraint.html#scipy.optimize.NonlinearConstraint) are good reads for this.

The cost is minimized given that some constraint inequality is satisfied (i.e. `(cstr(x) - epsilon) <= 0`). The cost and constraint functions are likely to check for very different things, so from an interpretability standpoint it does not make sense to combine them at this level. Also, they are combined into one cost function under the hood but this is the responsibility of the actual optimizer used, not the `ConstrainedMinimizer` class. Combining them here means that off the shelf constrained optimizers cannot be used.

While satisfying constraints does mean passing some boolean condition, treating them as discrete tests means we can't apply optimization tools that expect continuous and differentiable functions (which is an expectation of most constrained optimizers).

alonkukl: I see, so how does the user specifies the constraint inequality ? in the scipy [documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.NonlinearConstraint.html#scipy.optimize.NonlinearConstraint) there is an upper and a lower bound. Do you think there should one here as well?

mtweiden: That's a good point. It's not passed here because it fits better with how the `Minimizer.minimize` function call looks. In the current implementation, actual instances of the `ConstrainedMinimizer` have the success threshold as a parameter in `__init__`. There are different ways to [pass the success threshold given the optimizer](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.fmin_slsqp.html), so I think handling this internally still makes sense.

edyounis: If it makes a difference, I think keeping the instantiation API open for now is valuable. So if you think there is a modification to Minimizer API that makes better sense, we should discuss it.

Issue #262, opened 2024-07-17
RuntimeWarnings from `get_runtime().cancel(future)`


WolfLink:
I have some code of the form:

```
my_future = get_runtime().map(...)

result1 = get_runtime().next(my_future)
result2 = get_runtime().next(my_future)

get_runtime().cancel(my_future)
```

I get warning messages of the form:

```
myvenv/lib/python3.10/site-packages/bqskit/utils/cachedclass.py:45: RuntimeWarning: coroutine 'RuntimeTask.run' was never awaited
  hash_a = all(
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

The line `hash_a = all(` is a seemingly random line of code each time.  Note that the warnings don't show up every time; but they seem to show up more frequently when my CPU is under heavy load.

It seems to be related to `get_runtime().cancel(my_future)` because if I replace that line with `await my_future` I no longer see the warnings (but the code takes undersireably longer obviously).

WolfLink:
So I found the issue.  You are supposed to `close()` coroutines if you aren't going to `await` them.  (It's also safe to `close()` a coroutine after it has been `await`ed).  Neither happens when a task gets cancelled currently.

Relevant StackOverflow: https://stackoverflow.com/questions/62045387/how-to-suppress-coroutine-was-never-awaited-warning


I have two possible fixes to suggest:
1. The easier way is to simply implement `__del__` for `RuntimeTask` which will close its coroutine.
2. The more explicit way would be to go `close()` all relevant coroutines when cancel gets called.


The 1st approach is a lot simpler.  I'm pretty sure the warning happens when the unused coroutine gets garbage collected, so this approach is just letting python know it was ok for that coroutine to go unused.

The 2nd approach is a lot more work to implement, but it has a benfit in that if there is some future bug where a `RuntimeTask` falls through the cracks when it wasn't supposed to be cancelled, the warning might pop up to let you know.  The 1st approach will make it so *any* time a `RuntimeTask` is never executed for any reason, the warning will not appear.

PR #263, opened 2024-07-17
Implemented `__del__` to clean up hanging coroutines

WolfLink:
This is intended to be a simple fix for https://github.com/BQSKit/bqskit/issues/262

Currently, when `RuntimeTask`s are cancelled, they sometimes end up garbage collected with coroutines that are never `await`ed.  When those coroutines get garbage collected, `RuntimeWarning`s appear.

I've implemented `__del__` for `RuntimeTask` which simply `close()`s the coroutine, which lets Python know its ok that the coroutine was never used.

### DIFF
--- a/bqskit/runtime/task.py
+++ b/bqskit/runtime/task.py
@@ -104,6 +104,10 @@ class RuntimeTask:
     def cancel(self) -> None:
         """Ask the coroutine to gracefully exit."""
         if self.coro is not None:
+            # If this call to "close" raises a RuntimeError,
+            # it is likely a blanket try/accept catching the
+            # error used to stop the coroutine, preventing
+            # it from stopping correctly.
             self.coro.close()
         else:
             raise RuntimeError('Task was cancelled with None coroutine.')
--- a/bqskit/runtime/task.py
+++ b/bqskit/runtime/task.py
@@ -103,7 +103,10 @@ class RuntimeTask:
 
     def cancel(self) -> None:
         """Ask the coroutine to gracefully exit."""
-        self.coro.close()
+        if self.coro is not None:
+            self.coro.close()
+        else:
+            raise RuntimeError('Task was cancelled before its coroutine was started.')
 
     async def run(self) -> Any:
         """Task coroutine wrapper."""
--- a/bqskit/runtime/task.py
+++ b/bqskit/runtime/task.py
@@ -103,8 +103,7 @@ class RuntimeTask:
 
     def cancel(self) -> None:
         """Ask the coroutine to gracefully exit."""
-        if self.coro is not None:
-            self.coro.close()
+        self.coro.close()
 
     async def run(self) -> Any:
         """Task coroutine wrapper."""
--- a/bqskit/runtime/task.py
+++ b/bqskit/runtime/task.py
@@ -115,4 +115,3 @@ class RuntimeTask:
     def is_descendant_of(self, addr: RuntimeAddress) -> bool:
         """Return true if `addr` identifies a parent (or this) task."""
         return addr == self.return_address or addr in self.breadcrumbs
-
--- a/bqskit/runtime/task.py
+++ b/bqskit/runtime/task.py
@@ -101,6 +101,11 @@ class RuntimeTask:
         """Initialize the task."""
         self.coro = self.run()
 
+    def cancel(self) -> None:
+        """Ask the coroutine to gracefully exit."""
+        if self.coro is not None:
+            self.coro.close()
+
     async def run(self) -> Any:
         """Task coroutine wrapper."""
         if inspect.iscoroutinefunction(self.fnargs[0]):
@@ -111,7 +116,3 @@ class RuntimeTask:
         """Return true if `addr` identifies a parent (or this) task."""
         return addr == self.return_address or addr in self.breadcrumbs
 
-    def __del__(self) -> None:
-        """Close the coroutine to make sure it is not left hanging."""
-        if self.coro is not None:
-            self.coro.close()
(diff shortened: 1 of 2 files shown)


edyounis:
Thanks for doing this! Good catch on this. It seems like pre-commit is unhappy about something that Alon fixed and was recently merged into main. Can you merge main into this and make sure the pre-commit check passes. Let me know if you need help or want to chat. Thanks again for doing this!

edyounis:
Also, do you have any references for the close method on coroutine? I only found the [`generator.close()`](https://docs.python.org/3/reference/expressions.html#generator.close) documentation.

Going off the generator documentation, we may need to catch and throw some exceptions. Catch from within the __del__ method and throw from within the task execution, maybe.

WolfLink:
I don't think its totally unreasonable to deal with exceptions in `__del__`. [This documentation](https://docs.python.org/2/reference/datamodel.html#object.__del__) claims that "exceptions that occur during their execution are ignored, and a warning is printed to sys.stderr instead".  [This StackOverflow](https://stackoverflow.com/questions/13977970/ignore-exceptions-printed-to-stderr-in-del) claims that only applies to *uncaught* exceptions, and its possible to `try`/`except` as usual.'

Reading more about how `coroutine.cancel()` works, it looks like the way it tries to stop the coroutine is raising an exception inside the coroutine, and if that fails to cause the coroutine to exit (if the exception is caught), then `cancel()` will raise an exception.

This might happen if the coroutine gets cancelled while its in a `try` block followed by a blanket `except`, which is not an uncommon pattern, so this concern is more than just theoretical.


Ironically, one solution might be to wrap the call to `cancel()` in a blanket `try`/`except` block.

WolfLink:
Honestly I think we should not try to catch the exceptions that might come from `cancel`, and if anyone encounters it, the recommendation is to avoid blanket `except`statements, or at least to think carefully about when they do.


The remaining question then is if the behavior of these errors becoming warnings because of being called in `__del__` or if they should get raised as proper errors when the task is cancelled.

WolfLink:
I changed it to do explicit cancelling instead of using `__del__`.

bqskit/runtime/task.py
```diff
@@ -101,6 +101,11 @@ def start(self) -> None:
         """Initialize the task."""
         self.coro = self.run()
 
+    def cancel(self) -> None:
+        """Ask the coroutine to gracefully exit."""
+        if self.coro is not None:
+            self.coro.close()
```
edyounis, line 107: Is there ever an issue with a double-close? Should we throw an error here if the coro is None?

bqskit/runtime/worker.py
```diff
@@ -327,12 +330,17 @@ def _handle_cancel(self, addr: RuntimeAddress) -> None:
...
+                task.cancel()
         self._tasks = {
             a: t for a, t in self._tasks.items()
             if not t.is_descendant_of(addr)
         }
 
         # Remove all tasks that are children of `addr` from delayed tasks
+        for task in self._delayed_tasks:
+            if task.is_descendant_of(addr):
+                task.cancel()
```
edyounis, line 342: Delayed tasks don't have their coroutine started. This is safe to remove.

edyounis (changes requested):
Thanks for making this change. I do think this is a better method. Now we know if we see that warning that something is truly going wrong.

bqskit/runtime/task.py
WolfLink: Throwing an error if coro is `None` can be done simply by not checking if its `not None` and you will get something like "object of type NoneType does not have a method close()".

If you want to throw a more specific error we can do that.

When I had the check in `__del__` I had to not throw an error when `coro` is `None` because that seems to happen frequently.  I am new to the inner working of the bqskit runtime so I don't know what the conditions for `coro` being `None` are.

WolfLink: Based on my read of the python documentation, I think closing a coroutine multiple times shouldn't be an issue, but I'll test it.

WolfLink:
> Edit: Do you still think we should put a try-except around close()?

I think we shouldn't because an error indicates the coroutine didn't exit properly which is a problem the user should know about and suggests something should be fixed either in BQSKit or the user's code.

WolfLink:
I guess the ‘if coro is None` check needs to be there after all

bqskit/runtime/task.py
edyounis: Since this error is not something a user would be debugging we don't necessarily need to be more specific, but we can. A message like "Inactive RuntimeTask cannot be canceled." should be plenty for us to get some immediate insight.

I am curious, if you structure it as:

```python
if self.coro is None:
    raise RuntimeError("...")
self.coro.close()
```

If that error occurs in the usual cases, It shouldn't, right?

bqskit/runtime/worker.py
```diff
@@ -338,6 +347,10 @@ def _handle_cancel(self, addr: RuntimeAddress) -> None:
             if not t.is_descendant_of(addr)
         ]
 
+        # if there was an error earlier, raise it now
+        if error is not None:
+            raise error
```
edyounis, line 352: Why do you delay the raise? Shouldn't the system crash immediately, or is this to test something?

WolfLink: The idea is to finish cancelling tasks as much as possible before letting the system crash.  Although I suppose if its gonna crash, the system isn't going to be in a great state anyway.  Should I just let it crash immediately?

edyounis: Yeah, let it crash immediately.

WolfLink:
@edyounis This one should be good to merge now.  It includes all the tweaks we talked about.

PR #266, opened 2024-08-01
Adding sub_state function to  StateVector

jkalloor3:
This function returns a normalized state vector over a sub-selection of qudits in the Circuit.


bqskit/qis/state/state.py
```diff
@@ -166,6 +166,43 @@ def is_qudit_only(self, radix: int) -> bool:
...
+            return StateVector(self)
+
+        excluded_perm = [
+            x
+            for x in range(self.num_qudits)
+            if x not in location
+        ]
+        included_perm = list(location)
+
+
```
edyounis, line 191: Two empty lines

bqskit/qis/state/state.py
```diff
@@ -166,6 +166,43 @@ def is_qudit_only(self, radix: int) -> bool:
                 states.
         """
         return all([r == radix for r in self.radixes])
+    
+    def get_sub_state(self, location: CircuitLocationLike) -> StateVector:
+        """
+        Return the StateVector corresponding to the specified location
+        in the Circuit.
```
edyounis, line 173: I would drop the "in the Circuit" and make this one line. I am also curious how does renormalization happen? Can you provide either more detail about what this does mathematically, provide a reference, or add an example (or two)? Ideally, all of the above.

edyounis (changes requested):
Thanks for continuing to improve BQSKit and add features. This looks good, a couple of small changes are requested. Please, also add a test or two.

PR #269, opened 2024-08-14
Register `Workflow`s

mtweiden:
Given a `MachineModel` or `GateSet`, a custom `Workflow` can be registered to overwrite the default `Workflow` for a given `optimization_level` in `bqskit.compile`.

- Added `workflow_registry` object to maintain registered `Workflow`s
- Added `register_workflow` function so that custom `Workflow`s can be associated with a `MachineModel` or `GateSet` and `optimization_level`
- Modified `bqskit.compile` to check for registered `Workflow`s

TODO:
- [x] Add tests


bqskit/compiler/registry.py
```diff
@@ -0,0 +1,99 @@
...
+    """
+    Register a workflow for a given machine model.
+
+    Args:
+        machine_or_gateset (MachineModel | GateSetLike): A MachineModel or
+            GateSetLike to register the workflow under. If a circuit is
+            compiled targeting this machine or gate set, the registered
+            workflow will be used.
+
+        workflow (list[BasePass]): The workflow or list of passes that whill
```
edyounis, line 48: will*

bqskit/compiler/registry.py
```diff
@@ -0,0 +1,99 @@
...
+from bqskit.compiler.gateset import GateSetLike
+from bqskit.compiler.machine import MachineModel
+from bqskit.compiler.workflow import Workflow
+from bqskit.compiler.workflow import WorkflowLike
+from bqskit.ir.gate import Gate
+
+_logger = logging.getLogger(__name__)
+
+
+_workflow_registry: dict[MachineModel | GateSet, dict[int, Workflow]] = {}
```
edyounis, line 31: What's the design decision behind `MachineModel | GateSet` vs just `MachineModel`?

bqskit/compiler/registry.py
```diff
@@ -0,0 +1,99 @@
...
+            workflow will be used.
+
+        workflow (list[BasePass]): The workflow or list of passes that whill
+            be executed if the MachineModel in a call to `compile` matches
+            `machine`. If `machine` is already registered, a warning will be
+            logged.
+
+        optimization_level (Optional[int]): The optimization level with
+            which to register the workflow. If no level is provided, the
+            Workflow will be registered as level 1. (Default: 1)
```
edyounis, line 55: I am not sure there should be a default for this parameter. I think it should need to be specified here.

bqskit/compiler/registry.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Raises:
+        TypeError: If `machine_or_gateset` is not a MachineModel or GateSet.
+
+        TypeError: If `workflow` is not a list of BasePass objects.
```
edyounis, line 60: You typically don't need to call out TypeErrors since that is implicit in the parameter type annotations. You don't need to remove them tho.

bqskit/compiler/registry.py
```diff
@@ -0,0 +1,99 @@
...
+            m += f'GateSet, got {type(machine_or_gateset)}.'
+            raise TypeError(m)
+
+    workflow = Workflow(workflow)
+
+    for p in workflow:
+        if not isinstance(p, BasePass):
+            m = 'All elements of `workflow` must be BasePass objects. Got '
+            m += f'{type(p)}.'
+            raise TypeError(m)
```
edyounis, line 78: This check should be moved to the Workflow object with an `is_workflow` or similarly named static method.

bqskit/compiler/registry.py
```diff
@@ -0,0 +1,99 @@
...
+
+    global _workflow_registry
+    new_workflow = {optimization_level: workflow}
+    if machine_or_gateset in _workflow_registry:
+        if optimization_level in _workflow_registry[machine_or_gateset]:
+            m = f'Overwritting workflow for {machine_or_gateset} '
+            m += f'at level {optimization_level}.'
+            _logger.warn(m)
```
edyounis, line 86: Warnings should be made with the warning module here, see the table at the top of: https://docs.python.org/3/howto/logging.html

We might even want to suggest some tips for fixing this issue, e.g. uninstalling/unimporting conflicting packages?

bqskit/compiler/registry.py
```diff
@@ -0,0 +1,99 @@
...
+        _workflow_registry[machine_or_gateset].update(new_workflow)
+    else:
+        _workflow_registry[machine_or_gateset] = new_workflow
+
+
+def clear_registry() -> None:
```
edyounis, line 92: When will this be used?

bqskit/compiler/registry.py
```diff
@@ -0,0 +1,99 @@
...
+are indexed by optimization level. This object should not be accessed directly
+by the user, but instead through the `register_workflow` function.
+
+Example:
+    model_t = SpecificMachineModel(num_qudits, radixes)
+    workflow = [QuickPartitioner(3), NewFangledOptimization()]
+    register_workflow(model_t, workflow, level)
+    ...
+    new_circuit = compile(circuit, model_t, optimization_level=level)
+"""
```
edyounis, line 15: The module docstring will be less visible than the `register_workflow` documentation. You should migrate this helpful information to there.

bqskit/compiler/gateset.py
```diff
@@ -231,5 +231,9 @@ def __repr__(self) -> str:
         """Detailed representation of the GateSet."""
         return self._gates.__repr__().replace('frozenset', 'GateSet')
 
+    def __hash__(self) -> int:
+        """Hash of the GateSet."""
+        return self.__repr__().__hash__()
```
edyounis, line 236: Is this hash dependent on the order in which gates were appended to the set or some other internal ordering? That is, can two equal gate sets have different hashes with this method?

bqskit/compiler/compile.py
```diff
@@ -669,6 +671,18 @@ def build_workflow(
...
+    # optimization_level
+    for machine_or_gateset in _workflow_registry:
+        if isinstance(machine_or_gateset, GateSet):
+            gate_set = machine_or_gateset
+        else:
+            gate_set = machine_or_gateset.gate_set
+        gs_match = gate_set == model.gate_set
+        ol_found = optimization_level in _workflow_registry[machine_or_gateset]
+        if gs_match and ol_found:
+            return _workflow_registry[machine_or_gateset][optimization_level]
```
edyounis, line 684: I am not sure I follow this check. Why don't you simply check if `(model, optimization_level) in _workflow_registry`?

edyounis (changes requested):

bqskit/compiler/registry.py
mtweiden: Clearing the registry is needed for tests to pass. 

This function can be removed if you think it will encourage people to play around with things in a way that will break stuff.

bqskit/compiler/registry.py
mtweiden: `MachineModel` requires that you know beforehand how many qudits will be in some target, which is not something that is known when adding to the registry in namespace packages. Really what matters is just if the `GateSet` matches what is expected.

bqskit/compiler/compile.py
mtweiden: Supporting `GateSet` checking is more valuable than `MachineModel` checking because `GateSet` checking doesn't require knowing a qudit count for the target.

bqskit/compiler/registry.py
edyounis: This is true, but I think this feature's goal is entirely married to the MachineModel concept. By this, I mean an extension developer should be using this feature to automatically say: if you are compiling for this machine, this is a much better workflow than the off-the-shelf version

If you are sure your method is generic over a whole class of MachineModels, then you should have a `build_model` function exported with your package. At the end of this package, you can register the newly built model if its isn't registered yet.

Maybe I am missing something tho, lmk if you think there is a compelling use case for the GateSet. Either here or through Slack. My worry is with the potential conflicts/overlaps between GateSet and Model resolutions.

bqskit/compiler/registry.py
edyounis: Yeah, if it is for internal use only, add an `_` or remove it entirely. If motivated enough, people can always directly access the underlying data structure.

PR #272, opened 2024-09-03
Register workflows by target type

mtweiden:
This pull request replaces the `_compile_registry` object with type specific registries. This enables default workflows to be registered depending on the type of the target.

Removed
- `_compile_registry`

Added
- `_compile_circuit_registry`, `_compile_unitary_registry`, `_compile_stateprep_registry`, `_compile_statemap_registry`
- `registry_workflow` function now takes as a parameter a target type (`'circuit'`, `'unitary'`, `'stateprep'`, `'statemap'`)

Fixed
 - Bug in `bqskit/compiler/compile.py` which always sets the target as `Circuit(1)` for targets of type `UnitaryMatrix`.

### DIFF
--- a/bqskit/compiler/registry.py
+++ b/bqskit/compiler/registry.py
@@ -66,8 +66,7 @@ def register_workflow(
             `key`. If `key` is already registered, a warning will be logged.
 
         optimization_level (int): The optimization level with which to
-            register the workflow. If no level is provided, the Workflow will
-            be registered as level 1.
+            register the workflow.
 
         target_type (str): Register a workflow for targets of this type. Must
             be 'circuit', 'unitary', 'stateprep', or 'statemap'.
--- a/bqskit/compiler/registry.py
+++ b/bqskit/compiler/registry.py
@@ -8,13 +8,17 @@ from bqskit.compiler.workflow import Workflow
 from bqskit.compiler.workflow import WorkflowLike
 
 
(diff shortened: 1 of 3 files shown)


bqskit/compiler/registry.py
```diff
@@ -34,10 +38,13 @@ def register_workflow(
             be executed if the MachineModel in a call to `compile` matches
             `key`. If `key` is already registered, a warning will be logged.
 
-        optimization_level ptional[int): The optimization level with which
+        optimization_level (Optional[int]): The optimization level with which
             to register the workflow. If no level is provided, the Workflow
             will be registered as level 1.
```
edyounis, line 43: This says optional, but it isn't in the parameter list.

bqskit/compiler/registry.py
```diff
@@ -34,10 +38,13 @@ def register_workflow(
...
 
+        target_type (str): Register a workflow for targets of this type. Must
+            be 'circuit', 'unitary', 'stateprep', or 'statemap'.
```
edyounis, line 46: Should this default to circuit?

edyounis (changes requested):

bqskit/compiler/registry.py
mtweiden: I'm not opposed to this. I think if this happens, there should be a check to see if a `'circuit'` (or any other) flow has been registered so that a warning can be made just in case.

PR #282, opened 2024-09-23
Adding Multiplexed Rotation Gates

jkalloor3:
Pull request creates 2 new multiplexed rotation gates.

These gates have n -1 select qubits and a single target qubit. It has 2^(n-1) parameters corresponding to all of the possible configurations of the n - 1 select qubit. Each configuration of the n - 1 select qubits applies to a rotation gate (parameterized by the respective parameter) to the target qubit. The select qubits are unchanged.

These gates are used extensively in QSD and Block ZXZ Unitary Decomposition.


edyounis (changes requested):
Can you add the gates the documentation docstring in `bqskit.ir.gates.__init__`? Other than that, lgtm

PR #300, opened 2024-10-29
Fix `KeyError` for already popped tasks

### LINKED ISSUE #301: Race Condition in Worker._process_task_completion leading to a crash
## Description
There is a race condition between _process_task_completion and _handle_cancel methods in the Worker class, where concurrent modification of self._tasks can lead to an exception. The race condition occurs when one thread is processing task completion while another thread handles a cancellation request.

### Affected Methods

Method 1: _process_task_completion
```
def _process_task_completion(self, task: RuntimeTask, result: Any) -> None:
    assert task is self._active_task
    packaged_result = RuntimeResult(task.return_address, result, self._id)
    if task.return_address not in self._tasks:  # Check existence
        return
    # ... other operations ...
    self._tasks.pop(task.return_address)  # Attempt to remove

```
Method 2: _handle_cancel
This method runs in a separate thread and modifies the same self._tasks dictionary, causing the race condition.

### Race Condition Scenario

1. Thread 1 (in _process_task_completion): Checks if task.return_address not in self._tasks - condition is false (task exists)
2. Thread 2 (in _handle_cancel): Removes the task from self._tasks
3. Thread 1 continues: Attempts self._tasks.pop(task.return_address)
Raises KeyError because the task was already removed by the cancel handler

mtweiden:
Some more complex instantiation workflows may have several instances of tasks with the same `RuntimeAddress`. In these cases, multiple calls to remove the task associated with the same `RuntimeAddress` may call `Worker._tasks.pop` after it has already been popped. 

Adding a default pop value of `None` avoids raising a `KeyError` in these scenarios.

Fixes #301

### DIFF
--- a/bqskit/runtime/worker.py
+++ b/bqskit/runtime/worker.py
@@ -524,7 +524,7 @@ class Worker:
             self._conn.send((RuntimeMessage.RESULT, packaged_result))
 
         # Remove task
-        self._tasks.pop(task.return_address)
+        self._tasks.pop(task.return_address, None)
 
         # Cancel any open tasks
         for mailbox_id in self._active_task.owned_mailboxes:


alonkukl:
I think that the current fix of adding None as a default for the pop is a good fast workaround, but race conditions between the two threads should be treated more carefully.

I will let @edyounis decide if to accept this PR or not.

edyounis:
@alonkukl What would the benefit of a mutex be for this? The outcome of the race condition shouldn't matter as long as we avoid the error, right?

edyounis (approved):

alonkukl:
In the current code of the function the only issue I can see is that cancel
messages will be sent multiple times to sub tasks,  this doesn't seem clean
to me.

Regarding the _tasks data structure, I agree that it doesn't matter.

Alon

PR #324, opened 2025-05-28
Fix StateVector Zero and add test

### LINKED ISSUE #305: StateVector Zero Constructor needs radixes to be passed in
https://github.com/BQSKit/bqskit/blob/670658689bde03e4adda8d933455d2c6fbecf5a4/bqskit/qis/state/state.py#L203

This is a bug where the radixes parameter from the constructor is unused; it should be passed onto the object initializer. The radixes argument details the dimension of each qudit in the state vector.

WingCode:
closes #305

Added test

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,7 @@ repos:
     -   id: fix-encoding-pragma
         args: ['--remove']
 -   repo: https://github.com/s-weigand/docformatter
-    rev: 1ec30b7
+    rev: 886a032
     hooks:
     -   id: docformatter
         args:


edyounis:
Oh yeah, nice catch. I didn't look at the pre-commit log and thought it was an actual pre-commit found issue. Instead its an issue with the workflow. Would you mind making the proposed workflow change?

WingCode:
@edyounis I have bumped the cache version. I hope it solves the issue

WingCode:
@edyounis I taken the liberty of bumping up docformatter to the latest commit available on the repo. I hope it fixes the issue of missing commit ID for docformatter.

WingCode:
@edyounis Would it be possible to approve the pipeline again so that we could check if the pipeline runs successfully ?

PR #358, opened 2026-01-23
Improving `CouplingGraph.get_subgraphs_of_size`

jclark-ce:
Hello,

I was using ScanPartitioner to partition some circuits, and I noticed that it performs very slowly with increasing block size even on circuits with linear connectivity. After testing `CouplingGraph`'s `get_subgraphs_of_size`, I discovered that it takes a time exponential $k$ to produce the $\sim{n}$ groups of size $k$ in a linear coupling graph.

I have therefore written a small optimization which takes advantage of the fact that if the current group has been encountered before, no new groups can be found by continuing to search. The change reduces the time to find all groups of size $8$ on a 10x10 grid coupling graph from $\sim{88}$ seconds down to $\sim{2}$ seconds.

The changed version saves all qubit groups to `locations` as `frozenset` and only recurs if the new group has not been encountered before. Groups with less than `size` qubits are dropped before returning from `get_subgraphs_of_size`. The change does slightly increase the memory complexity of the method since all groups of size $\leq{k}$ are saved, although in most cases this is not an appreciable increase (since the number of groups with size $<k$ is usually less than the number of group with size $=k$).

Because groups with size $<k$ are produced for free with this method, it might also be sensible to add an interface which allows them to be left in for use cases like ScanPartitioner which need them anyway.


edyounis (dismissed):

jclark-ce:
Okay, I updated `get_subgraphs_of_size` to use a hybrid of [Redelmeier](https://page.mi.fu-berlin.de/rote/Kram/redelmeier.pdf)'s algorithm and ESU (since, from testing, Redelmeier is faster with only a tiny bit more memory usage). I also added the `allow_smaller` flag to `CouplingGraph.get_subgraphs_of_size` and `MachineModel.get_locations` so that applications like ScanPartitioner can use it and updated `ScanPartitioner.calculate_qudit_groups` to take advantage of the flag.

Regarding contributing TDAG to BQSKit, I may be interested in doing that, I'll send you an email in the next few days to discuss it.

Issue #360, opened 2026-02-04
Circuit and gate testing strategies frequently fail


jclark-ce:
After some tests were run for a PR (#358), I noticed that the `test_remove_atomic_units_removes_num_mq_gates` test in `test_seed.py` frequently fails due to a hypothesis unsatisfiable error (it does this even on the main branch, at least on my machine; not sure if this is a known issue/if it has come up before). It seems that the error is caused by the `gates` testing strategy frequently failing to find a gate meeting the testing requirements, which causes the `circuit` strategy to restart. The likelihood of this happening increases the longer the desired test circuit (i.e. almost no circuits with >=3 gates are generated). 

https://github.com/BQSKit/bqskit/blob/172b48c305917c7b763e0e2928a74cafad163520/bqskit/utils/test/strategies.py#L240-L257

This can be partially mitigated by switching the `assume` statements to `.filter`, but still fails occasionally. Filtering the gates directly (i.e. without using hypothesis to filter) seems to work very reliably, but I suppose might be more prone to missing test cases if not written well. The `frozen_gates` method also sometimes returns a completely frozen (i.e. constant) gate even when `constant` is `False`, which contributes (slightly) to the issue.

I am open to making a second PR to fix this issue, but wasn't sure what solution would be preferred, or even if this issue is being caused by some other issue (e.g. hypothesis behaves somewhat differently on my machine).

edyounis:
Thank you for your concern. The whole test suite is flaky at this point and soon to be replaced. If it isn't too much trouble to open this PR, then let's try it out on the Github Actions, otherwise, don't worry about it.

PR #362, opened 2026-02-11
Fix testing strategies

jclark-ce:
Address #360, changes are: 
- Made the `gates` strategy filter gates ahead of time rather than relying on hypothesis
- Fixed `frozen_gates` to return only non-constant gates when requested, as well as accounting for None `constant` inputs
- Switched `floats` calls to use the default `width` argument, which drastically reduces the rejection rate for param lists (this defaults to 64-bit floats, which should produce valid params as far as I can tell)

### DIFF
--- a/bqskit/ir/gates/__init__.py
+++ b/bqskit/ir/gates/__init__.py
@@ -43,6 +43,8 @@ BQSKit Gates (:mod:`bqskit.ir.gates`)
     SwapGate
     SqrtXGate
     SXGate
+    SqrtXdgGate
+    SXdgGate
     SycamoreGate
     TGate
     TdgGate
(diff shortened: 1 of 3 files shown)


jclark-ce:
I guess I must've forgotten to re-run the tests before making the `floats` call change. Anyway, the errors in the last test were due to nan popping up in the results with large param values (which caused the tests to fail since nan is not equal to itself). The new version clamps the floats calls to the range $[-2^{16},2^{16}]$, which is roughly equivalent to using 16-bit floats (but without the failure rate). 

Having actually run tests on the new version, it looks like they all pass with the exception of `test_power_gate`, which might be an actual bug, either in `PowerGate` or in the `test_power_gate` fixture.

PR #375, opened 2026-05-18
Add citations feature

kbidlack:
Closes #368; adds way for users to retrieve BibTeX citations for the algorithms used during compilation, making it easier to cite individual works rather than just the catch-all BQSKit citation.

## Changes:
Added a new Citation dataclass (keyed by DOI for deduplication) and `@cite` class decorator for attaching citation metadata. BasePass objects now have a `get_citations()` method which returns `set[Citation]`. Workflow objects have both `get_citations() -> set[Citation]` and `gather_citations() -> dict[Citation, list[BasePass]]`, the latter of which maps citations to the passes that use them.
I also annotated `GeneralizedSabreAlgorithm` with `@cite` for its references.


README.md
```diff
@@ -44,8 +44,8 @@ You can use the [software disclosure](https://doi.org/10.11578/dc.20210603.2)
 to cite the BQSKit package.
 
 Additionally, if you used or extended a specific algorithm, you should cite
-that individually. BQSKit passes will include a relevant reference in
-their documentation.
+that individually. Citations can be programatically retrieved for passes and
+workflows via `get_citations` and/or `gather_citations`.
```
edyounis: Let's make this comment an "as well as" and keep the reference both in the documentation and in the `get_citations` method. Granted, we should only enforce the programatic way going forward, and it is up to contributors to include the reference in the documentation. Maybe, change "BQSKit passes will" to "BQSKit passes should"

edyounis: We should give an extremely simple example on how to use `get_citations` here.

bqskit/compiler/workflow.py
```diff
@@ -66,6 +68,19 @@ def __init__(self, passes: WorkflowLike, name: str = '') -> None:
         if len(self._passes) == 0:
             raise ValueError('Expected at least one pass in workflow.')
 
+    def get_citations(self) -> set[Citation]:
+        """Return the set of citations for this pass and all its parent
+        classes."""
```
edyounis: Every docstring's first line should fit in one line. If you do a single-line docstring, ensure it can fit completely on one line. Otherwise, do a multi-line docstring, where the first line is a summary that completes in one line.

FYI, not the most important note in the world, but I figured I'd mention. [PEP 257](https://peps.python.org/pep-0257/) is a good read.

An easy/lazy change (what I would do) is change "set of citations" to "citation set" or change it completely to "Return all citations associated with this workflow." 

BTW, you say "for this pass and all its parent classes" here in workflow, but not in `BasePass`.

bqskit/compiler/workflow.py
```diff
@@ -66,6 +68,19 @@ def __init__(self, passes: WorkflowLike, name: str = '') -> None:
...
+        return set(self.gather_citations().keys())
+
+    def gather_citations(self) -> dict[Citation, list[BasePass]]:
+        """Return a mapping of citations to the passes that use them."""
+        result: defaultdict[Citation, list[BasePass]] = defaultdict(list)
+        for pass_ in self._passes:
+            for citation in pass_.get_citations():
+                result[citation].append(pass_)
```
edyounis: Some of the pass objects may be quite large. Is all of that information necessary for the purpose of this method? Maybe we can just return `dict[Citation, list["PassName"]`?

edyounis (changes requested):
I really like that you made the citations a decorator. I made some smaller comments throughout, but one larger change I would like to see is to only request the DOI for the cite. That is, `@cite('10.1145/3445814.3446718')` should be a complete citation. Few reasons for this:

1) Lessens the buy-in from developers. We want to encourage the developers to use this feature. This extra step could be the dealbreaker. (This is true, as unfortunate as it is)

2) You can programmatically go from DOI to bibtex:
```python
import urllib.request

doi = "10.1145/3445814.3446718"
url = f"https://doi.org/{doi}"

req = urllib.request.Request(url)
req.add_header('Accept', 'application/x-bibtex')

with urllib.request.urlopen(req) as response:
    bibtex_entry = response.read().decode('utf-8')

print(bibtex_entry)
```

3) Makes the code look prettier :), The big block of text at the beginning of a pass is unsightly to read.

4) In all honesty, a DOI is all you need. You can copy-paste it into scholar and get an accurate result for everything in our domain.

README.md
edyounis: Can you also go through all of the passes that have a reference in them, and add the DOI cite, so this feature can be completely used by all the passes in BQSKit?

bqskit/compiler/workflow.py
kbidlack: > BTW, you say "for this pass and all its parent classes" here in workflow, but not in `BasePass`.

Right, I forgot to ask you about this—`get_citations` exists on `BasePass`, so it gets inherited by `Workflow`. Because of that, I added `gather_citations` for `Workflow` objects to separate the functions; one returns the mapping and one returns just a set of `Citation` objects (or DOI strings, soon). 

Given that, I added this note and function to override the inherited one. Do you think it should be done differently?

edyounis: I think the way you implemented this is good. I was referring to the docstring. The docstring should be rewritten as we talked about in our last meeting.

edyounis (dismissed):
LGTM.

A minor note regarding `gather_citations`: If you say you return a dict, return a dict; if you return a defaultdict, say you return a defaultdict. This nit pick isn't worth not approving the PR, but I wanted to bring this to your attention.

kbidlack:
> LGTM.
> 
> A minor note regarding `gather_citations`: If you say you return a dict, return a dict; if you return a defaultdict, say you return a defaultdict. This nit pick isn't worth not approving the PR, but I wanted to bring this to your attention.

Addressed

kbidlack:
@jkalloor3 pre-commit fail is unrelated to my pull request, but I can try to fix it if you'd like

jkalloor3:
> @jkalloor3 pre-commit fail is unrelated to my pull request, but I can try to fix it if you'd like

Yeah, that would be great if you can do in a separate PR first. That way we just make sure future PRs don't break things

kbidlack:
> > @jkalloor3 pre-commit fail is unrelated to my pull request, but I can try to fix it if you'd like
> 
> Yeah, that would be great if you can do in a separate PR first. That way we just make sure future PRs don't break things

Fixed!

jkalloor3 (approved):

PR #377, opened 2026-07-02
Fix CI

kbidlack:
Fixes two issues:

1. Mypy was complaining that `os.sched_setaffinity` is not available on macOS, but this never failed in CI since the runner is Linux. I updated to check `hasattr(os, 'sched_setaffinity')` instead.
```pytb
bqskit/runtime/worker.py:777: error: Module has no attribute
"sched_setaffinity"  [attr-defined]
            os.sched_setaffinity(0, [cpu])
            ^~~~~~~~~~~~~~~~~~~~
```

2.
```pytb
bqskit/exec/runners/quest.py:133: error: Argument 3 to "RunnerResults" has
incompatible type "ndarray[Any, Any]"; expected "Sequence[float]"  [arg-type]
    ...      return RunnerResults(circuit.num_qudits, circuit.radixes, probs)
                                                                       ^~~~~
Found 2 errors in 2 files (checked 513 source files)
```

Fixed by changing `probs` to `probs.to_list()`

### DIFF
--- a/bqskit/exec/runners/quest.py
+++ b/bqskit/exec/runners/quest.py
@@ -130,7 +130,11 @@ class QuestRunner(CircuitRunner):
         # 5. Average and return results
         probs = np.sum(np.array([result.probs for result in results]), axis=0)
         probs /= self.sample_size
-        return RunnerResults(circuit.num_qudits, circuit.radixes, probs)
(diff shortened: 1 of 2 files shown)


jkalloor3 (approved):
LGTM

PR #378, opened 2026-07-18
OpenQudit Migration Part 1: Gates, pt. 1

kbidlack:
This PR begins the migration of BQSKit's `Gate` class to `OpenQudit`'s `UnitaryExpression` via a (temporary?) `Gate._expr`, removing the need for many previously manually-implemented methods.

Every gate in `bqskit/ir/gates/` (constant and parameterized) now optionally carries a module-level `_expr: UnitaryExpression` attribute. Where a gate sets `_expr`, the base `Gate` class automatically provides `num_params`, `num_qudits`, `radixes`, `dim`, `get_unitary`, `get_grad`, and `get_unitary_and_grad`.

### Changes

The minimum supported Python version has been updated to **3.10**, as `OpenQudit` does not support any versions below that.

#### `bqskit/ir/gate.py`; `bqskit/qis/unitary/differentiable.py`; `bqskit/ir/gates/qubitgate.py` and `.../quritgate.py`

* Added optional `_expr: UnitaryExpression` annotation.
* Removed `DifferentiableUnitary`; all methods have been promoted to `Gate`.
* Properties `name`, `num_params`, `num_qudits`, `radixes`, `dim`, and functions `get_unitary`, `get_grad`, and `get_unitary_and_grad` prefer `_expr`.
* Removed `QubitGate` and `QutritGate` which existed only to add a `radixes` property. `Gate` now derives this automatically from `_expr`.

#### `bqskit/ir/gates/constantgate.py`

`ConstantGate` no longer inherits from `DifferentiableUnitary` and no longer implements `get_unitary` or `get_unitary_and_grad` (those moved to `Gate`). `get_grad` is kept on `ConstantGate`, returning `np.array([])`, since all constant gates are differentiable with a zero gradient.

#### `bqskit/ir/gates/composedgate.py`

`ComposedGate.is_differentiable()` now checks `gate.is_differentiable()` instead of `isinstance(gate, DifferentiableUnitary)`.

### Tests

* `tests/qis/unitary/test_meta.py`: `TestIsDifferentiable` suite removed
* `tests/ir/circuit/test_properties.py`: `test_value` in `TestIsDifferentiable` removed; it tested `isinstance(gate,   DifferentiableUnitary)`.
* `tests/ir/gates/test_native.py`:`test_get_grad` and `test_get_unitary_and_grad` now mark `CRYGate` as `xfail` (due to a pre-existing `bqskitrs`).
* `tests/ir/gates/composed/test_frozenparams.py`: Hypothesis float strategy for `test_u3_rx` is now bounded to `[1e-15, 1e15]` to avoid the case where extreme (e.g. >1e307; <1e-307) angles fail the test.


bqskit/ir/gates/constant/permutation.py
```diff
@@ -34,11 +40,30 @@ def __init__(
             raise ValueError('Expected positive integer, got %d' % num_qudits)
 
         self._num_qudits = num_qudits
+        self._radixes = tuple([2] * num_qudits)
         self.location = tuple(location)
         self._utry = PermutationMatrix.from_qubit_location(
             num_qudits, self.location,
```
jkalloor3: Shouldn't we delete this?

bqskit/ir/gates/constant/pd.py
```diff
@@ -70,3 +71,15 @@ def __init__(self, index: int, radix: int = 3) -> None:
         omega = np.exp(2j * np.pi * index / radix)
         diags = [(-omega ** 2) if i == index else 1 for i in range(radix)]
         self._utry = UnitaryMatrix(np.diag(diags), self.radixes)
```
jkalloor3: Delete this and get from expression

bqskit/ir/gates/constant/itoffoli.py
```diff
@@ -33,6 +36,9 @@ class IToffoliGate(ConstantGate, QubitGate):
 
     _num_qudits = 3
     _qasm_name = 'iccx'
+    _expr = _Controlled(
+        _UnitaryExpression('iX() { [[0,i],[i,0]] }'), [2, 2],
+    )
     _utry = UnitaryMatrix(
```
jkalloor3: Should derive from expression

bqskit/ir/gates/constant/s.py
```diff
@@ -20,6 +22,10 @@ class SGate(ConstantGate, QubitGate):
 
     _num_qudits = 1
     _qasm_name = 's'
+    # See YGate for why `_utry` stays a hand-written exact matrix instead
```
jkalloor3: Remove this comment since the YGate comment is gone

bqskit/ir/gates/constant/sdg.py
```diff
@@ -20,6 +23,10 @@ class SdgGate(ConstantGate, QubitGate):
 
     _num_qudits = 1
     _qasm_name = 'sdg'
+    # See YGate for why `_utry` stays a hand-written exact matrix instead
```
jkalloor3: Remove

bqskit/ir/gate.py
```diff
@@ -12,22 +12,138 @@
...
+                UnitaryMatrix: The unitary matrix, see
+                :func:`Unitary.get_unitary` for more info.
+
+                np.ndarray: The unitary's gradient, see :func:`get_grad`.
+
+        Notes:
+            Can be overridden to potentially speed up optimization by
+            calculating both at the same time.
+        """
+        return (self.get_unitary(params), self.get_grad(params))
```
jkalloor3, line 137: Does OpenQudit have a get_unitary_and_grad function? Would be slightly faster than calling each separately

bqskit/ir/gate.py
```diff
@@ -12,22 +12,138 @@
...
+
+    def is_differentiable(self) -> bool:
+        """Return true if this gate has a gradient definition."""
+        return (
+            type(self).get_grad is not Gate.get_grad
```
jkalloor3: I may be inclined to only check if it is an OpenQudit expression, I don't think we have to support gates that aren't expressions anymore

tests/ir/gates/test_native.py
```diff
@@ -66,7 +66,17 @@ def test_get_unitary(gate: Gate) -> None:
     assert np.allclose(circ.get_unitary(x), circuit.get_unitary(x))
 
 
-@pytest.mark.parametrize('gate', NATIVE_GATES, ids=lambda gate: repr(gate))
+NATIVE_XFAIL_CRY = [
+    pytest.param(
+        gate, marks=pytest.mark.xfail(
+            reason='bqskitrs incorrectly computes gradient for CRYGate',
```
jkalloor3, line 72: Do we not use OpenQudit to compute this gradient now?

bqskit/ir/gate.py
kbidlack: Not that I'm aware of

tests/ir/gates/test_native.py
kbidlack: We do. The test does both though.
It first tests BQSKit's implementation:

```py
from bqskit.ir import Circuit as Circ
...
def test_get_grad(gate: Gate) -> None:
    size = gate.num_qudits
    circ = Circ(size, radixes=gate.radixes)
    circ.append_gate(gate, list(range(size)))
    num_params = circ.num_params
    x = np.random.random((num_params,))
```

Then it tests `bqskitrs`:
```py
    circuit = Circuit(circ)
    grad_python = circ.get_grad(x)
    grad_rust = circuit.get_grad(x)
    for py, rs in zip(grad_python, grad_rust):
        assert np.allclose(py, rs)
```

Originally, BQSKit's gradient implementation was incorrect (it seems `bqskitrs`'s took from BQSKit), but now that we just directly call OpenQudit, it's correct, so the test fails.

tests/ir/circuit/test_properties.py
```diff
@@ -1054,16 +1053,6 @@ class TestIsDifferentiable:
     def test_type(self, r6_qudit_circuit: Circuit) -> None:
         assert isinstance(r6_qudit_circuit.is_differentiable(), bool)
 
-    def test_value(self, gate: Gate) -> None:
-        circuit = Circuit(gate.num_qudits, gate.radixes)
-        assert circuit.is_differentiable()
```
jkalloor3, line 1059: Bring the test back outside of DifferentiableUnitary

tests/ir/gates/composed/test_frozenparams.py
```diff
@@ -10,8 +10,16 @@
...
+# 8.988465674311579e30 = max double size / 2
+@given(
+    floats(
+        min_value=-1e15, max_value=1e15,
+        allow_nan=False, allow_infinity=False,
+    ),
+)
 def test_u3_rx(angle: float) -> None:
     frozen_u3 = FrozenParameterGate(U3Gate(), {1: -np.pi / 2, 2: np.pi / 2})
+    print(angle)
```
jkalloor3: Remove

bqskit/utils/test/strategies.py
```diff
@@ -139,8 +139,12 @@ def state_likes(
     return draw(sampled_from([vec, vec.numpy]))
 
 
+_GATE_BASE_CLASSES = {'ComposedGate', 'QuditGate', 'GeneralGate'}
```
jkalloor3, line 142: What is this change again?

jkalloor3 (approved):
LGTM!!

PR #380, opened 2026-07-31
Fix ZXZXZDecomposition doubling synthesis for diagonal unitaries

stevenhofmeyr:
When the middle rotation is Clifford (t == 0 or +-pi), the SX.RZ(t).SX block commutes with the outer RZ/U1 gates, so only their sum or difference is actually determined by the target unitary -- how that total is split between the two outer gates is a free gauge choice. The unconstrained formulas split it evenly, turning one gate that needs synthesis into two. This is the common case: any diagonal single-qubit unitary, e.g. every run of Z-axis rotations, hits it.

Collapse the gauge instead so the whole rotation lands on one outer gate and the other becomes an exact identity (angle 0).

### DIFF
--- a/bqskit/passes/rules/zxzxz.py
+++ b/bqskit/passes/rules/zxzxz.py
@@ -13,6 +13,12 @@ from bqskit.ir.gates.parameterized.rx import RXGate
 from bqskit.ir.gates.parameterized.rz import RZGate
 from bqskit.ir.gates.parameterized.u1 import U1Gate
 
+# Tolerance for recognizing the ZXZXZ middle rotation angle as Clifford (0 or
+# +-pi). This only has to separate those two values from every other angle
+# the decomposition produces, and both come out exact to floating-point
+# precision, so this can be tight.
+_CLIFFORD_ANGLE_TOL = 1e-10
+
 
 class ZXZXZDecomposition(BasePass):
     """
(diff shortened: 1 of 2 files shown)


bqskit/passes/rules/zxzxz.py
```diff
@@ -13,6 +13,12 @@
 from bqskit.ir.gates.parameterized.rz import RZGate
 from bqskit.ir.gates.parameterized.u1 import U1Gate
 
+# Tolerance for recognizing the ZXZXZ middle rotation angle as Clifford (0 or
+# +-pi). This only has to separate those two values from every other angle
+# the decomposition produces, and both come out exact to floating-point
+# precision, so this can be tight.
+_CLIFFORD_ANGLE_TOL = 1e-10
```
jkalloor3, line 20: I am not a huge fan of hard-setting this tolerance, I am going to rework how error is handled so we can update based on that. For now, maybe we can have this as an argument to the pass (and have the default value be 1e-10)?

jkalloor3 (changes requested):

bqskit/passes/rules/zxzxz.py
WolfLink: I’d be a bit worried about having this tolerance there by default because it might result in approximation happening in a part of the workflow where that isn’t expected.


To expand on this a bit: The use case of synthesizing as U3, and then decomposing to ZXZXZ, may result in gates that are close but not quite the case where this applies. This code will end up rounding those gates. I don't think that should be the default behavior.  **`ZXZXZDecomposition` should be as exact as possible by default.**  Errors on the scale of `1e-10` do add up in large circuits.

